PR #12, opened 2026-03-11
get positions with `ndam`

cattabiani:
## Context

Fix: #13

## Scope

- use `ndam` to get the position files. Do not rely on `report.h5` or 1 ts simulation anymore
- simplify `getPositions`. We use `ndam` gid splitting and save in different files. This may temporarily increase the stress of the file system (before we were saving in various folders). However, since the saving of the positions to file is going to be erased in the future I do not want to overcomplicate something that is going to go away anyway
- add cli call to `get_positions`
- reshuffle `examples` folder since a dedicated `getPositions` folder and scripts are no longer necessary
- leave the `run_get_positions.py` script since I did not handle some folders in `examples` like `hippocampus`
- update various readmes

## Note

Given the recent [fix in neurodamus](https://github.com/openbraininstitute/neurodamus/pull/499), one of the tests in here will fail if you try to run the pytest unit tests with an old version of neurodamus. use 4.2.0


bluerecording/getPositions.py
```diff
@@ -425,35 +440,18 @@ def checkAxonsFirst(morphology):
...
-
-    if len(ids) == 0:
-        return 1
-
-    data = getMinimalReport(report,ids)
-
-    colIdx = data.columns # node_id and Section IDs for each cell
-    cols = np.array(list(data.columns))
+    rSim = bp.Simulation(path_to_simconfig)
+    population = rSim.circuit.nodes[population_name]
```
WeinaJi: Can we get the node population directly via libsonata ? Bluepysnap uses libsonata anyway behind the scenes. That would remove one dependency.

WeinaJi (commented):
1. Can you add a mpi test for the `get_positions` function in tests/unit-mpi ? 
2. The naming convention in this repository mixes snake_case and camelCase. It needs to be updated to follow standard Python conventions.

WeinaJi (approved):
One nitpick, in the PR's title and description, "get positions from neurodamus" means to get the cells' discretizations right? To me "positions" means the coordinates of cell segments, but this PR doesn't get them from neurodamus. You can change the title at merging time.

cattabiani:
@WeinaJi technically we get the positions of the compartments here:

```
        cols = np.array([
            (p.gid, s)
            for p in points
            for s in sorted(p.sclst_ids)
        ], dtype=np.int64)

        return ids, cols, node_manager.population_name
   ```
   
   Ok, there is also other stuff but that was the main point. It developed to become more like a discretization so, ok. I will change that

PR #16, opened 2026-03-16
use ndam also for roto-translation and reconstruction

cattabiani:


cattabiani:
found a different way to do this

PR #17, opened 2026-03-17
use `ndam` to roto-translate

cattabiani:
## Context

Fix: #18 #23

## Scope

- Replace ad-hoc scipy quaternion rotation (float64) with neurodamus's local_to_global_coord_mapping (float32). Joe said [here](https://github.com/openbraininstitute/BlueRecording/issues/15) that this is ok. In this case, results will differ slightly due to the different numerical precision in rotation
- Removes duplicated rotation logic and several helper functions
- Axon reconstruction and soma position convention unchanged — only rotation precision differs (max ~0.00025 µm)
- Rework `getPositions`. This should be just cosmetic (easier to read now) + a few improvements for robustness. It should not change results
- remove duplicate points. The main reason we were rotating in float64 (I suspect)
- remove dendFirst logical path. Obsolete. #22 and #23 for more info


jamesgking (commented):
I see that the instructions for examples/circuitTest/Readme.txt don't work anymore. Were you planning to keep this and update the instructions on a later commit?

jamesgking (approved):

PR #31, opened 2026-03-27
Standardize project structure, add single-cell example, clean up tests

cattabiani:
## Context

The project had accumulated non-PEP 8 naming conventions (camelCase module and test file names), fragile test fixtures sharing mutable state via session-scoped temp files, and the `single_cell_l5_tpc` example lacked simulation/analysis documentation. The weight computation code was split across two modules (`writeH5.py` and `writeH5_prelim.py`) with no clear reason for the separation.

Fix: #33 #32  #29

## Scope

- Bump `requires-python` from `>=3.9` to `>=3.10` in `pyproject.toml`
- `single_cell_l5_tpc` example:
  - Add `simulation_config_distant.json`
  - Add simulation and analysis sections to README (neurodamus instructions, Figure 2 reference with verification warning)
  - Fix `make_figures.ipynb`: switch to `%matplotlib inline`, add `plt.show()`, remove `savefig`
- Base `README.md`: add pointer to single-cell example, update module references from `writeH5`/`writeH5_prelim` to `weights`
- Merge `bluerecording/writeH5.py` + `bluerecording/writeH5_prelim.py` into `bluerecording/weights.py`, update all imports (cli, tests, example scripts)
- Delete old `writeH5.py` and `writeH5_prelim.py`
- Rename test files to PEP 8 snake_case:
  - `test_writeH5.py` + `test_h5_initialization_SONATA.py` → `test_weights.py`
  - `test_get_positions.py` → `test_positions.py` (unit + MPI)
  - `test_single_cell_get_positions.py` → `test_single_cell_positions.py`
  - `test_h5py_MPI.py` → `test_h5py.py`
- Rework test fixtures:
  - Replace shared mutable session/module-scoped fixtures with helper functions in `tests/helpers.py`
  - Tests now use `tmp_path` for isolated file I/O — no ordering dependencies
  - Slim `conftest.py` down to path constants only
- Add `ci` mode to `run_tests.sh` (`./run_tests.sh ci`) mirroring GitHub Actions (`-m "not skip_in_ci"`)

## Tests

All unit and MPI tests pass locally:
- `./run_tests.sh ci` — fast CI-like suite (excludes `skip_in_ci` tests)
- `./run_tests.sh` — full suite including slow data-dependent integration tests

No test logic was changed — only fixture wiring, file names, and import paths.

PR #36, opened 2026-04-09
Bump version to 0.3.0

cattabiani:
## Context

Bump version to `0.3.0` to mark the first OBI-era release.

This goes on top of #31, which brings the codebase to a stable point after the BBP to OBI migration. Key changes since `v0.2.1`:

- Neurodamus integration for discretization and roto-translation via `ndam`. Greatly streamlined weights creation
- Consolidated weight computation into a single `weights` module
- MPI-parallel test coverage with CI
- Standardized examples with a ready-to-run `single_cell_l5_tpc` setup
- PEP 8 naming, isolated test fixtures, and updated documentation
- `setup.sh` now builds `NEURON` from source with `libsonatareport` support (pinned commit), enabling end-to-end simulation out of the box
- Added `--no-cache` flag to setup.sh for a clean rebuild of the entire environment (preserves downloaded data)
- Added `--quick` flag to setup.sh to skip `NEURON`/`libsonatareport`/`neurodamus-models builds` (used by CI)
Lazy import of neurodamus in `circuit.py` so lightweight installs don't break at import time. It also skips `neurodamus` install (clarified in docs)
- Switched to `hatch-vcs` for dynamic versioning from git tags
- Bumped NEURON pin to `9.0.1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,18 @@ source setup.sh --dev
 
 If you want to skip the system packages installation append `--no-system` to the previous lines.
 
+Use `--quick` to skip building NEURON, libsonatareport, and neurodamus-models from source (useful for CI or when you only need the Python package):
+
+```bash
+source setup.sh --dev --quick
+```
+
+Use `--no-cache` to wipe the virtual environment and all build artifacts before reinstalling from scratch (downloaded data is preserved):
+
+```bash
+source setup.sh --dev --no-cache
+```
+
 Finally, if you want to run the full testing suite you need `--data`. See the [Testing](#testing) section for details.
 
 
@@ -48,8 +60,10 @@ The initial input data of `BlueRecording` includes a [compartment report](https:
 On the first run, the script will:
 1. Create a Python virtual environment with MPI-enabled `h5py` and `mpi4py`
 2. Clone and build `libsonatareport`
-3. Clone and build `neurodamus-models` (neocortex)
-4. Install the `bluerecording` package
+3. Clone and build NEURON from source (with `libsonatareport` support)
+4. Install `neurodamus` from source
+5. Clone and build `neurodamus-models` (neocortex)
+6. Install the `bluerecording` package
 
 In subsequent sessions, running `source setup.sh` again will simply activate the existing environment.
 
--- a/README.md
+++ b/README.md
@@ -115,6 +115,12 @@ pytest -v -m "not skip_in_ci" tests/unit
 
 This is also what runs in CI, where we avoid downloading the full test data to keep pipelines fast and lightweight.
 
+> **Note:** If you installed with the `--quick` flag (i.e. `source setup.sh --dev --quick`), NEURON, neurodamus, and libsonatareport are not available. In that case only the non-MPI unit tests can run:
+> ```bash
+> pytest -v -m "not skip_in_ci" tests/unit
+> ```
+> The MPI tests and any test that requires neurodamus will be skipped or fail.
+
 To run only the slow, data-intensive tests (e.g., single cell and 100-cell integration tests):
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ source setup.sh --dev
 
 If you want to skip the system packages installation append `--no-system` to the previous lines.
 
-Use `--quick` to skip building NEURON, libsonatareport, and neurodamus-models from source (useful for CI or when you only need the Python package):
+Use `--quick` to skip building NEURON, libsonatareport, neurodamus, and neurodamus-models from source (useful for CI or when you only need the Python package):
 
 ```bash
 source setup.sh --dev --quick
@@ -115,7 +115,7 @@ pytest -v -m "not skip_in_ci" tests/unit
 
 This is also what runs in CI, where we avoid downloading the full test data to keep pipelines fast and lightweight.
 
-> **Note:** If you installed with the `--quick` flag (i.e. `source setup.sh --dev --quick`), NEURON, neurodamus, and libsonatareport are not available. In that case only the non-MPI unit tests can run:
+> **Note:** If you installed with the `--quick` flag (i.e. `source setup.sh --dev --quick`), NEURON, neurodamus, libsonatareport, and neurodamus-models are not available. In that case only the non-MPI unit tests can run:
 > ```bash
 > pytest -v -m "not skip_in_ci" tests/unit
 > ```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "bluerecording"
-version = "0.3.0"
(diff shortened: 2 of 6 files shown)


WeinaJi:
Why not 1.0.0 as there are so many big changes, and it is going to be used by OBI ?

README.md
jamesgking, line 109: minor: should some of the **Testing** section highlight that only `pytest -v -m "not skip_in_ci" tests/unit` can be done when the `--quick` instal option was used?

cattabiani: mm yes. I can add that

cattabiani:
>Why not 1.0.0 as there are so many big changes, and it is going to be used by OBI ?

@WeinaJi I guess it is mostly personal preference. The code is not in perfect condition yet. it is just usable. Anyway, if you want and everybody is ok, I can move it to 1.0.0

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "bluerecording"
-version = "0.2.1"
+version = "0.3.0"
```
WeinaJi: Can you use dynamic versioning instead ?

You can copy the neurodamus way of using `hatch-vcs` in https://github.com/openbraininstitute/neurodamus/blob/main/pyproject.toml#L48

WeinaJi (commented):
That's also fine for me if you prefer to tag 1.0.0 later. Up to you. 
One comment about using dynamic versioning:

setup.sh
```diff
@@ -1,9 +1,20 @@
 #!/usr/bin/env bash
 # setup.sh — Full MPI-enabled Python environment setup
 
+# -------------------------
+# Pinned versions (edit here to update)
+# -------------------------
+NEURON_COMMIT="0d990513b"
```
jplanasc: Can we maybe pin NEURON versions instead of commits?

IIrc, on AWS and Azure we use NEURON 9.0.1 (which includes your commit).

setup.sh
```diff
@@ -13,13 +24,17 @@ for arg in "$@"; do
...
+        --quick) QUICK=1 ;;
         --help|-h)
             echo "Usage: source setup.sh [OPTIONS]"
             echo ""
             echo "Options:"
             echo "  --dev         Install development version (includes -e pip install)"
             echo "  --no-system   Skip system package installation"
             echo "  --data        Download and unpack datasets"
+            echo "  --no-cache    Remove venv, cloned repos, and build artifacts (keeps data)"
+            echo "  --quick       Skip NEURON, libsonatareport, and neurodamus-models builds"
```
jplanasc: Are you missing neurodamus here (and below)?

If I understood correctly, you only want neurodamus in the case you import NEURON as well?

setup.sh
cattabiani: I am following the neurodamus way. I am pinning with a commit as it is in neurodamus

setup.sh
cattabiani: with quick I do not need neuron ... not sure about neurodamus. it is really just for the ci. for testing base functions

setup.sh
jplanasc: Hhmm... I don't see it in Neurodamus repo?

https://github.com/openbraininstitute/neurodamus/blob/e5abd33ad2cc5a4450c2d5190e32afc11ee385d8/pyproject.toml#L32

setup.sh
jplanasc: But this is what you're doing right? You're just installing neurodamus when `--quick` is not passed?

https://github.com/openbraininstitute/BlueRecording/blob/339ed5ad8626c8ae4447eddd4991d18da773bcdc/setup.sh#L167-L170

I'm just saying that you should also mention neurodamus in the help.

setup.sh
```diff
@@ -118,13 +159,17 @@ else
...
     pip install --no-cache-dir --no-binary=h5py h5py --no-build-isolation
-
-    pip install git+https://github.com/openbraininstitute/neurodamus.git@main
 fi
 
+# -------------------------
+# Install libsonatareport, NEURON, neurodamus, neurodamus-models
+# (skipped with --quick)
+# -------------------------
+if [[ $QUICK -eq 0 ]]; then
```
jplanasc, line 171: I think this whole block inside the `if` should be indented, for better readability?

setup.sh
cattabiani: It is here

https://github.com/openbraininstitute/neurodamus/blob/e5abd33ad2cc5a4450c2d5190e32afc11ee385d8/.github/workflows/simulation_test.yml#L36

setup.sh
cattabiani: true, the help does not mention it. corrected

setup.sh
jplanasc: You added this in neurodamus, too 🤣 

In that case, I'd strongly suggest it be changed in both repos, we've tested NEURON 9.0.1 and we know it's working properly with our tools.

However, I let it to @cattabiani and @WeinaJi to make the final decision.

WeinaJi: In neurodamus CI, we have `NEURON_BRANCH` that could point to a release.  Sure, we could set it to `9.0.1` now. 
And I would prefer to keep the option `NEURON_COMMIT_ID` in the neurodamus CI because sometimes we need to develop in NEURON and test in Neurodamus. 
Here in BlueRecording, yes, we could simply use the latest NEURON release.

pyproject.toml
cattabiani: done

setup.sh
cattabiani: done

setup.sh
cattabiani: ok, done

setup.sh
```diff
@@ -118,44 +159,87 @@ else
...
+            -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR
+
+        cmake --build nrn/build --parallel
+        cmake --build nrn/build --target install
+    fi
+
+    # -------------------------
+    # Install neurodamus
+    # -------------------------
+    pip install git+https://github.com/openbraininstitute/neurodamus.git@main
```
jplanasc, line 220: Up to you, but installing from the latest commit may bring in instability to your CI.

I think we make frequent enough releases in neurodamus to pin a specific release, but again, I leave it up to you.

jplanasc (commented):
I think there's one comment to resolve. From my side, once addressed, looks good to me.

setup.sh
cattabiani: It happened already that I needed my own changes here. That is the reason I am getting main. For example the latest release that I did doesn't have some needed changes (I forgot which exactly). Thus, I would need to pin to a commit anyway now. For now I would leave it as is

jplanasc: Fair point, I wanted to point it out just in case.

jplanasc (approved):
LGTM

PR #37, opened 2026-04-09
Replace bluepysnap with libsonata

WeinaJi:
`bluepysnap` uses libsonata under the hood. For the main purpose of this repo, i.e. generating the weights file, `libsonata` is sufficient. So This PR moves `bluepysnap` to the optional dependency. 
It is still required for analysis code in examples and notebooks where it enables users to analyze simulation reports in greater detail.


bluerecording/circuit.py
```diff
@@ -54,8 +55,7 @@ def init_circuit(path_to_simconfig: str):
         dtype=np.int64,
     ).reshape(-1, 2)
 
-    r_sim = bp.Simulation(path_to_simconfig)
     population_name = node_manager.population_name
-    population = r_sim.circuit.nodes[population_name]
+    population = get_node_population(path_to_simconfig, population_name)
```
cattabiani: since the function is used only here and is 2 lines of code I would prefer to simply inline it

cattabiani (commented):
I have just 1 comment

bluerecording/circuit.py
WeinaJi: The script in examples and notebooks require the same functionality, currently via `bluepysnap`. 
If you are ok to keep bluepysnap for the examples and notebooks, this function would be only called here, yes, I could inline it. If to remove bluepysnap in examples and notebooks later, a separate function is better.

cattabiani: I will eventually replace bluepysnap in examples and notebooks but this will come later. In the mean time it is less code to handle. I am also considering that I have an update to utils and that function may slightly interphere. Anyway, it is non-blocking. We can also merge as-is. As a personal preference I think inlined is simpler to reason about

WeinaJi: Done, inlined.

cattabiani (approved):
LGTM

PR #44, opened 2026-04-14
update `utils`

cattabiani:
## Context
The code needs to be standardized and modernized. Here we focus in `utils.py`.

## Scope
- Removed `process_writeH5_inputs / process_subsampling_inputs` (dead sys.argv parsers)
- Inlined `process_subsampling_inputs` into `weights.py`
- Removed `concretize_path` (buggy `str.rstrip`-based path resolution), replaced all call sites with `pathlib.Path.resolve()`
- Renamed `getCircuitPath` → `get_circuit_path` with proper docstring and type hints
- Moved `getAtlasInfo` and `alignmentInfo` out of the library into the dead example script that uses them


bluerecording/positions.py
```diff
@@ -429,7 +429,7 @@ def get_morph_path(population, i, path_to_simconfig):
 
         finalmorphpath = remove_variables(js, finalmorphpath)
 
-        finalmorphpath = concretize_path(circuitpath,finalmorphpath) # Goes from relative to absolute path
+        finalmorphpath = str((Path(circuitpath).parent / finalmorphpath).resolve())
```
WeinaJi: For the whole block for finalmorphpath,  I think libsonata does resolve manifest for us. Neurodamus simple uses `node_prop.morphologies_dir`.

bluerecording/utils.py
```diff
@@ -1,186 +1,24 @@
...
+        Absolute path to the circuit configuration file.
+    """
     with open(path_to_simconfig) as f:
-
         circuitpath = json.load(f)['network']
     
-    circuitpath =  concretize_path(path_to_simconfig, circuitpath)
-    
-
+    circuitpath = str((Path(path_to_simconfig).parent / circuitpath).resolve())
```
WeinaJi: Can be done by libsonata.
```    
sim_conf = libsonata.SimulationConfig.from_file(path_to_simconfig)
return sim_conf.network
```
BTW, in `circuit.py`, the function `init_circuit()` also needs the circuit file path. This function can be used there.

bluerecording/positions.py
cattabiani: somehow I was expecting this to not be the case since Joe was doing all this parsing. My bad. Added

bluerecording/utils.py
cattabiani: I think I fixed this too. Same reasoning as before. Added

WeinaJi (approved):

examples/makeCsvFiles/writeNeuropixelsToCSV.py
```diff
@@ -1,10 +1,75 @@
...
+
+    return regionList, layerList
+
+def alignmentInfo(path_to_simconfig,target):
+
+    '''
+    Gets loction and angle information in order to align a probe with long axis of of the specified target (typically a cortical column)
+    '''
+    
+    population = getPopulationObject(path_to_simconfig)
```
WeinaJi, line 59: This function is broken because `getPopulationObject` is not defined any more. But it is only used by an example script, so you can fix/remove that later.

cattabiani: things in examples that are not single_cell or 100_cells are still a WIP. I have an issue to remind me of that

Issue #45, opened 2026-04-15
`MutableMorph` should be reworked and use composition


cattabiani:
`MutableMorph` has several design issues:

1. **Blind attribute copying.** The constructor iterates over `dir(morphioMorph)` and copies every attribute via `setattr`. This is fragile — it silently picks up internal/private attributes, and there's no way to know at a glance what the object's interface actually is. If MorphIO adds or renames an attribute, the behavior changes without any explicit code change on our side.

2. **Unnecessary mutability.** The class exists to make the morphio object "mutable", but the only mutation that ever happens is a single `m.points = cell.local_to_global_coord_mapping(m.points)` call in `get_morphology`. After that, the object is never mutated again. The entire class is built around enabling one assignment.

3. **Repeated low-level indexing.** Accessing section points requires `m.points[m.indices[sec.id]]` — a pattern that appears ~10 times across `get_axon_points` and `get_cell_positions`. This is noisy, error-prone, and obscures the intent.

PR #46, opened 2026-04-15
update `positions.py`

cattabiani:
## Context

Fix: #21 #45 #52

## Scope

This is the ff merge of 2 previously distinct prs. Let me know if this size it too big. Since they both insisted on the same file, I thought it was ok to merge them

### MutableMorph

Replaces the `MutableMorph` class (which copied all attributes from an immutable MorphIO object via `setattr` loop) with `PositionedMorphology`, a composition-based wrapper.

- `PositionedMorphology` holds the immutable morphio object internally and delegates `.sections` to it
- Coordinate transform (rotation + translation) is applied once at construction time via an optional `transform` callable, removing the need for post-construction mutation
- `indices` is lazily computed on first access
- New `section_points(sec_id)` method replaces the repeated `m.points[m.indices[sec.id]]` pattern
- Updated `get_morphology`, `get_axon_points`, `get_cell_positions` and all tests

### Update Positions

- Removed `remove_variables`, `tryFileNames`, `get_morph_path`, and `get_morphology` 
- replaced manual JSON parsing and manifest variable substitution with `libsonata.CircuitConfig.node_population_properties().morphologies_dir`
- Removed `MutableMorph` class, replaced by `PositionedMorphology` (immutable MorphIO wrapper with optional coordinate transform)
- Removed `concretize_path` from utils, inlined with `pathlib.Path.resolve()`
- Removed dead `sys.argv` parsing helpers from utils, inlined `processSubsampling` into weights
- Removed unused imports (`json`, `os`, `warnings`, `from .utils import *`, `from .circuit import init_circuit`)
- Removed global `warnings.filterwarnings('error', '', RuntimeWarning)`
- `get_axon_points`: extracted into `_get_cumulative_length`, `_get_branch_section_ids`, `_find_best_axon_branch`, `_collect_branch_points`, `_extrapolate_branch` — lazy memoized cumulative lengths
- `interp_points`: proper docstring, type hints, `np.column_stack` instead of hstack loop
- `interp_points_axon`: docstring, type hints, snake_case params, removed unused `soma_pos` parameter, build interpolators once via `np.column_stack`
- `_get_cell_positions` (was `get_cell_positions`): made private, type hints, docstring, `np.tile` for soma duplication, single `gid_mask`
- `getNewIndex` → `get_new_index`: snake_case, type hints
- `PositionedMorphology`: added type hints and docstrings to all properties and methods
- `utils.py`: snake_case, type hints, moved dead code to example script
- type hinting, doc strings and adoption of snake_case

### DIFF
--- a/bluerecording/positions.py
+++ b/bluerecording/positions.py
@@ -393,7 +393,7 @@ def interp_points_axon(
 def get_new_index(cols: np.ndarray) -> pd.MultiIndex:
     """Build a new MultiIndex by duplicating certain (id, section) column tuples.
 
-    Each column is kept once.  Non-somatic columns (section != 0) are
+    Each column is kept once. Non-somatic columns (section != 0) are
     duplicated when the next column tuple differs, to represent the end
     point of that section.  The last column is always repeated.
 
@@ -458,18 +458,18 @@ def _get_cell_positions(
     num_somas = np.sum(gid_mask & (cols[:, 1] == 0))
     xyz = np.tile(soma_pos.reshape(3, 1), num_somas)
 
-    for sec_name in sections[1:]:
-        num_compartments = np.sum(gid_mask & (cols[:, 1] == sec_name))
+    for sec_id in sections[1:]:
+        num_compartments = np.sum(gid_mask & (cols[:, 1] == sec_id))
 
-        if sec_name < 3 and replace_axons:
-            seg_pos = interp_points_axon(axon_points, running_lens, sec_name, num_compartments)
+        if sec_id < 3 and replace_axons:
+            seg_pos = interp_points_axon(axon_points, running_lens, sec_id, num_compartments)
         else:
-            sec_id = sec_name - 1
-            if sec_id >= m.num_sections:
+            morpho_sec_id = sec_id - 1
+            if morpho_sec_id >= m.num_sections:
                 # Beyond morphology sections → myelinated AIS
-                seg_pos = interp_points_axon(axon_points, running_lens, sec_name, num_compartments)
+                seg_pos = interp_points_axon(axon_points, running_lens, sec_id, num_compartments)
             else:
-                sec_pts = np.array(m.section_points(sec_id))
+                sec_pts = np.array(m.section_points(morpho_sec_id))
                 seg_pos = interp_points(sec_pts, num_compartments)
 
         xyz = np.hstack((xyz, seg_pos.T))
--- a/bluerecording/positions.py
+++ b/bluerecording/positions.py
@@ -361,10 +361,7 @@ def interp_points_axon(
         sec_len = 30
         start, end = 30, 60
 
-        idx = np.intersect1d(
-            np.where(running_lens <= end),
-            np.where(running_lens >= start),
-        )
+        idx = np.where((running_lens >= start) & (running_lens <= end))[0]
         axon_relevant = axon_points[idx]
         lens_relevant = (running_lens[idx] - start) / sec_len
 
--- a/bluerecording/positions.py
+++ b/bluerecording/positions.py
@@ -306,36 +306,22 @@ def get_axon_points(m: PositionedMorphology, center: np.ndarray) -> tuple[np.nda
 def interp_points_axon(
     axon_points: np.ndarray,
     running_lens: np.ndarray,
-    sec_name: int,
+    sec_id: int,
     num_compartments: int,
 ) -> np.ndarray:
     """Interpolate segment boundary points for a simulated-axon section.
 
-    The simulated axon has three sections identified by *sec_name*:
-
-    * 1 — first AIS section  (0–30 µm)
-    * 2 — second AIS section (30–60 µm)
-    * 3+ — myelinated section (60–1060 µm)
-
-    These are global section IDs originating from NEURON's section
-    numbering.  Neurodamus assigns them as offsets into the
-    ``SECTION_TYPES`` ordering (soma, axon, dend, apic, ais, node,
-    myelin).  After axon replacement the first two axon slots (indices 1
-    and 2) hold the AIS stubs.  Because the mapping from ID to type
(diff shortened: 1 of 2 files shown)


bluerecording/positions.py
```diff
@@ -1,79 +1,105 @@
...
+                self._indices.append(list(range(idx, idx + n)))
+                idx += n
+        return self._indices
 
-        for attr in dir(morphioMorph):
-            if '__' not in attr:
-                setattr(self,attr,getattr(morphioMorph,attr))
+    def section_points(self, sec_id: int) -> np.ndarray:
+        """Return the (possibly transformed) points for a given section."""
+        return self._points[self.indices[sec_id]]
```
WeinaJi: Is self.indices only used here? If so, maybe we can just store the section offset corresponding to self._points. So that we just need save one value per section instead of len(sec.points) ?

bluerecording/positions.py
```diff
@@ -272,285 +298,189 @@ def get_axon_points(m: MutableMorph, center: np.ndarray) -> tuple[np.ndarray, np
...
+    """Interpolate segment boundary points for a simulated-axon section.
 
-    if secName == 1: # First AIS section
+    The simulated axon has three sections identified by *sec_name*:
 
-        secLen = 30 # By construction, has length of 30 um
-        segLen = secLen / numCompartments # Assumes each segment has the same length
+    * 1 — first AIS section  (0–30 µm)
+    * 2 — second AIS section (30–60 µm)
+    * 3+ — myelinated section (60–1060 µm)
```
WeinaJi: How is the sec_name determined? Is it possible to use an enum type ? Is it more understandable than integers.

bluerecording/positions.py
```diff
@@ -272,285 +298,189 @@ def get_axon_points(m: MutableMorph, center: np.ndarray) -> tuple[np.ndarray, np
...
-        if len(axonRelevant) < 2: # If there aren't enough points, we estimate
+    elif sec_name == 2:  # Second AIS section (30–60 µm)
+        sec_len = 30
+        start, end = 30, 60
 
-            idxSmall = np.argmin(np.abs(runningLens - startPoint)) # Index closest to 30 um
+        idx = np.intersect1d(
+            np.where(running_lens <= end),
+            np.where(running_lens >= start),
+        )
```
WeinaJi: AI says it can be combined in a single `np.where`. I haven't tested it. You can have a try.
```
idx = np.where((running_lens <= end) & (running_lens >= start))[0]
```

bluerecording/positions.py
cattabiani: indices is used in several places beyond section_points: the axon branch helpers (_get_cumulative_length, _collect_branch_points) and _get_cell_positions all index into m.points via m.indices[sec.id]. 

In any case the suggestion makes sense as it would save some space and time. I will go ahead and do it

bluerecording/positions.py
cattabiani: These are global section IDs from NEURON's numbering: positional offsets into the SECTION_TYPES ordering, not categorical types. The same integer maps to different section types depending on per-cell section counts, so a static IntEnum wouldn't work. I expanded the docstring to explain why they're plain ints.

bluerecording/positions.py
cattabiani: mmm yes. added

bluerecording/positions.py
```diff
@@ -270,246 +303,194 @@ def get_axon_points(m: MutableMorph, center: np.ndarray) -> tuple[np.ndarray, np
...
+    """Build a new MultiIndex by duplicating certain (id, section) column tuples.
 
-def get_morphology(
-    population: libsonata.NodePopulation,
-    i: int,
-    morphologies_dir: str,
-    cell,
-) -> tuple[MutableMorph, np.ndarray]:
-    """Load and transform a morphology into circuit (global) coordinates.
+    Each column is kept once.  Non-somatic columns (section != 0) are
```
jplanasc: Are the double spaces after dot intentional here and in line 412?

jplanasc (dismissed):
LGTM

bluerecording/positions.py
WeinaJi: I see, so they are global section ids not section names, `sec_name` is a bit misleading. Can we rename it to `sec_id` ? 
And I think the new docstring can be shortened. BlueRecording users will not understand the Neurodamus details. Simply:
```
These are global section IDs after axon replacement.
```

cattabiani: done. I let you check the new docstring

bluerecording/positions.py
```diff
@@ -270,246 +303,180 @@ def get_axon_points(m: MutableMorph, center: np.ndarray) -> tuple[np.ndarray, np
...
 
-    if num_somas > 1: # If there is more than one somatic segment, we assume that they all have the same position
-        for k in np.arange(1,num_somas):
-            xyz = np.hstack((xyz,soma_pos.reshape(3,1)))
+    # Build soma columns — all somatic segments share the same position
+    num_somas = np.sum(gid_mask & (cols[:, 1] == 0))
+    xyz = np.tile(soma_pos.reshape(3, 1), num_somas)
 
-    for sec_name in list(sections[1:]):
+    for sec_name in sections[1:]:
```
WeinaJi: It seems better to rename sec_name -> sec_id in this part as well. What do you think ?
But what is `sec_id = sec_name - 1` below ?

bluerecording/positions.py
cattabiani: fixed

bluerecording/positions.py
cattabiani: the - 1 converts from the global NEURON section ID to the MorphIO section index.

Ok, for the renaming. Done

WeinaJi (approved):
LGTM

WeinaJi (approved):
Approve again.

PR #47, opened 2026-04-16
fix and refactor `get_axon_points`

### LINKED ISSUE #48: `get_axon_points` always picks the last section
In `get_axon_points`, `longestLength` was reset to 0 inside the per-leaf loop, so the 'find longest branch' fallback always picked the last leaf visited instead of the actual longest one. Move the initialisation before the loop so it accumulates correctly across all axonal leaves.

Only affects morphologies where no axonal branch reaches 1060 µm (the extrapolation path). Reference data regenerated accordingly.

We should fix that asap and regenerate the tests. It is also a good moment to refactor the function in general. It has:

- multi-purpose. It is not (S)OLID
- it explores a graph recomputing every time the length. memoization can help
- camelCase -> snake_case
- update docstring
- type hinting
- etc..

cattabiani:
## Context

Fix: #48 #21

## Scope

- Fix `longestLength` bug (#48): was reset inside the per-leaf loop, so the fallback always picked the last branch instead of the longest. Regenerated reference data.
- Decompose `get_axon_points` into 4 focused helpers: `_get_cumulative_length`, `_find_best_axon_branch`, `_get_branch_section_ids`, `_collect_branch_points`, `_extrapolate_branch`.
- Replace per-leaf full-path recomputation with lazy memoized cumulative lengths — each section is measured at most once.
- Flatten `soma_pos` upfront in `_collect_branch_points` to remove per-point shape-mismatch branch.
- Replace `np.vstack` in loop with list collection + single vstack.
- camelCase → snake_case across all `get_axon_points` helpers.
- Add type hints and docstrings to all new functions.
- Remove magic default target_length=1060 — callers must pass explicitly.


bluerecording/positions.py
```diff
@@ -72,125 +72,200 @@ def interp_points(coords, ncomps):
...
+            if length > longest_length:
+                longest_length = length
+                best_idx = idxs
+            if length > target_length:
+                break
 
-            if length > longestLength: # If this branch is longer than the previous longest branch we update the longest brancg
-                longestLength = length
-                bestIdx = idxs
+    need_extension = False
```
jamesgking, line 160: Thanks for finding this issue and fixing.

Do we have morphologies in the tests for both conditions need_extension true, need_extension false?

If not, maybe we can add in this or future pr

cattabiani: there are already. In `make_morphology` the axon is long enough (`need_extension` is `False`). `make_morphology_short` checks the other path

jamesgking: I'll take a closer look at those tests; I would have liked then to have triggered on the bad resetting of the longest_length variable

cattabiani: it was not detected because all the unit tests had just 1 branch. 1 branch -> longest branch == last branch

bluerecording/positions.py
```diff
@@ -72,125 +72,200 @@ def interp_points(coords, ncomps):
...
+    Returns:
+        points: (N, 3) array of 3D positions along the branch.
+        running_len: List of cumulative arc lengths, one per point.
+        current_len: Final cumulative arc length.
+    """
+    last_pt = soma_pos.flatten()
 
-    for x in idxs: # We iterate through the selected axonal sections
+    point_list = [last_pt.reshape(1, 3)]
+    running_len = [0]
```
WeinaJi, line 192: why are `point_list` and `running_len` initialized with the `last point`? Does the `section_ids` contain soma section? Would it add the soma points 2 times ?

bluerecording/positions.py
```diff
@@ -72,125 +72,200 @@ def interp_points(coords, ncomps):
...
-            runningLen.append(currentLen)
-
-
-            points = np.vstack((points, pt))
+        for pt in pts:
+            current_len += np.linalg.norm(pt - last_pt)
+            running_len.append(current_len)
+            point_list.append(pt.reshape(1, 3))
+            last_pt = pt
+            if current_len > target_length:
```
WeinaJi, line 204: The `break` is after list `append`. Should the last point > target_length be included in the results?

bluerecording/positions.py
cattabiani: No duplication. `section_ids` only contains neurite (axon) sections — the soma is never a MorphIO section (see the comment on line 38). The soma point is added once here as a starting anchor so the arc-length calculation correctly includes the gap between the soma center and the first axon point.

bluerecording/positions.py
cattabiani: Yes, it's intentional. The downstream consumer `interp_points_axon` interpolates positions along the branch at specific arc lengths up to `target_length`. To interpolate accurately at exactly `target_length`, you need a point beyond it... otherwise you'd be extrapolating past your last known point. Including that overshoot point ensures the interpolation has data on both sides of the target

jamesgking (dismissed):
LGTM

bluerecording/positions.py
WeinaJi: Ok, I got it wrong from the docstring of `_get_branch_section_ids`. It says soma->tip, maybe better to clarify that.

bluerecording/positions.py
WeinaJi: ok.

WeinaJi (dismissed):

bluerecording/positions.py
cattabiani: good point. I fixed the docstring so it is clear that the soma is not included

cattabiani:
I think I addressed all the comments. I let you check one last time if you have any last-minute comments

WeinaJi (approved):
LGTM!

Issue #51, opened 2026-04-20
remove libsonatareport from bluerecording


cattabiani:
Currently, we need `libsonatareport` only for `neurodamus-models`. We do not really need it. We should find a way to avoid passing it

We should also improve the installation procedure now that we talked and I have a clearer picture on what the environment on the platform will be like

PR #56, opened 2026-04-21
improve setup (no libsonatareport)

cattabiani:
## Context

Partial fix of #51

`libsonatareport` is still needed if you want to run the simulations

More details in [this discussion](https://github.com/openbraininstitute/neurodamus-models/issues/13#issuecomment-4289266009)

# Scope

- Declared most dependencies in `pyproject.toml` — `mpi4py`, `h5py`, `neurodamus`, `pandas`, `morphio`, `libsonata`, `voxcell`, `scikit-learn` are now proper base dependencies. `neuron` is an optional extra (`pip install bluerecording[neuron]`) because it may be built from source.

- Simplified `setup.sh` to a single mode — no more `--light` / `--dev` / `--full` flags. The script always builds the full simulation stack from source (libsonatareport → NEURON → neurodamus-models with reporting). This is the only use case for `setup.sh` — platform installs just use `pip install bluerecording[neuron]` or `pip install bluerecording`.

- `setup.sh` now works both sourced and executed:
  - `source setup.sh` — activates the environment in the current shell (as before)
  - `./setup.sh` — installs everything, writes `env.sh`, tells the user to `source env.sh`

- `setup.sh` detects active venvs via `$VIRTUAL_ENV` instead of checking for a `./venv` folder. If a venv is already active, it installs into that one. Pip installs always run (pip skips satisfied deps).

- `__init__.py` checks only what pip can't guarantee: h5py MPI support and neuron availability. Everything else is a declared dependency.

- CI (`ci.yml`) no longer calls `setup.sh`. It installs deps directly with pip (including neuron from wheel), builds neurodamus-models without reporting, and caches the result. Faster and simpler.

- Renamed `--no-cache` to `--clean-install`. Pinned neurodamus and neuron commits at the top of `setup.sh` (`NEURON_COMMIT`, `NEURODAMUS_COMMIT`).

- `mpi4py` installs from PyPI wheel instead of building from source. `neurodamus-models` clones with `--depth=1`.

- Updated README and `run_tests.sh` to reflect the single-mode setup.


bluerecording/__init__.py
```diff
@@ -3,11 +3,69 @@
...
+    the README for installation instructions.
+    """
+    try:
+        from mpi4py import MPI  # noqa: F401
+    except ImportError:
+        raise ImportError(
+            "bluerecording requires mpi4py built against your system's MPI library.\n"
+            "It is not listed in pyproject.toml because it must be compiled with MPI support.\n"
+            "Install it with: pip install --no-binary=mpi4py mpi4py\n"
+            "Or use 'source setup.sh' which handles this automatically."
```
mgeplf: why `source setup.sh` rather than running `setup.sh`? Are there environment variables that need to be propagated to my shell?

bluerecording/__init__.py
```diff
@@ -3,11 +3,69 @@
...
-if not h5py.get_config().mpi:
-    raise RuntimeError("h5py lacks MPI support. Reinstall it with MPI support.")
+def _check_dependencies():
+    """Verify that runtime dependencies not declared in pyproject.toml are available.
 
-if MPI.COMM_WORLD.size < 1:
-    raise RuntimeError("MPI not initialized correctly.")
\ No newline at end of file
+    bluerecording requires mpi4py, h5py (with MPI support), NEURON, and neurodamus.
+    These are not listed in pyproject.toml because they need special build flags
```
mgeplf: I think they can be listed in pyproject, this `_check_dependencies()` will give hints as to what needs to be done, I would say

bluerecording/__init__.py
```diff
@@ -3,11 +3,69 @@
...
+            "Install it with: pip install neuron\n"
+            "Or use 'source setup.sh' which handles this automatically."
+        )
+
+    try:
+        import neurodamus  # noqa: F401
+    except ImportError:
+        raise ImportError(
+            "bluerecording requires neurodamus.\n"
+            "It is not listed in pyproject.toml because it may need to be pinned to a\n"
```
mgeplf: `may need to be pinned to`; I think if you're going to say it `may` need, it's worth describing that somewhere, and pointing this comment to that description.

README.md
```diff
@@ -10,32 +10,46 @@ This branch provides code that produces an electrodes file compatible with the [
 
 ### Dependencies
 
-BlueRecording requires `mpi4py`, `h5py`, and `hdf5` built with MPI support. The provided `setup.sh` script handles all of this automatically, including building NEURON, neurodamus, and libsonatareport from source.
+BlueRecording requires several packages that are **not listed in `pyproject.toml`** because they need special build flags or may be built from source:
 
-`setup.sh` works for both macOS (with `brew`) and Linux (with `apt`) systems:
+| Package | Why it's not in pyproject.toml |
+|---|---|
+| `mpi4py` | Must be compiled against the system MPI library |
```
mgeplf: can the mpich/openmpi wheels that mpi4py publish be used:
https://mpi4py.readthedocs.io/en/stable/install.html#:~:text=The%20mpi4py%20project%20publishes%20MPICH%20wheels%20and%20Open%20MPI%20wheels ?

setup.sh
```diff
@@ -1,40 +1,50 @@
 #!/usr/bin/env bash
-# setup.sh — Full MPI-enabled Python environment setup
+# setup.sh — MPI-enabled Python environment setup for BlueRecording
 
 # -------------------------
 # Pinned versions (edit here to update)
 # -------------------------
-NEURON_COMMIT="9.0.1"
+NEURON_VERSION="9.0.1"
```
mgeplf: are only released versions supported?
the 9 series took a long time to get released, and the reporting things weren't included in 8, iirc, so requiring a released version might be infeasible.

setup.sh
```diff
@@ -1,40 +1,50 @@
...
+            echo "                Use for development and running the test suite."
+            echo "  --full        Build everything from source: libsonatareport, NEURON,"
+            echo "                neurodamus, and neurodamus-models with reporting enabled."
+            echo "                Editable install with test dependencies."
+            echo "                Required to run simulations with SONATA report generation."
+            echo ""
+            echo "Other options:"
+            echo "  --no-system   Skip system package installation (brew/apt)"
+            echo "  --data        Download and unpack example datasets (atlas, networks, FEM)"
             echo "  --no-cache    Remove venv, cloned repos, and build artifacts (keeps data)"
```
mgeplf: Not sure I follow what `--no-cache` does; isn't it more "remove install"?

setup.sh
```diff
@@ -213,40 +229,52 @@ if [[ $QUICK -eq 0 ]]; then
...
-
-        cmake --build neurodamus-models/build
-        cmake --install neurodamus-models/build
+# -------------------------
+# neurodamus-models
+#   full:  with reporting (linked against libsonatareport)
+#   light: without reporting
+# -------------------------
+if [ ! -d "neurodamus-models" ]; then
+    git clone https://github.com/openbraininstitute/neurodamus-models.git
```
mgeplf: does one need the full history, or can one only clone it with `--depth 1`?

bluerecording/__init__.py
cattabiani: yes, various neurodamus/neuron env variables. In addition the python virtual environment

bluerecording/__init__.py
cattabiani: listed as a comment: they are already there. Listed as a dependency: I decided to not do that. What is worse about these 2 cases:
- install in the wrong way the library (i.e. h5py)
- missing the library

The latter for me is clearer

setup.sh
cattabiani: this pinning is exactly what @jplanasc suggested. I let you fight... I will go fetch some /popcorn

setup.sh
cattabiani: yes. it is mainly for me. I need to verify often clean installations. Do you prefer another name?

setup.sh
cattabiani: true

README.md
cattabiani: I did not know there were wheels for this. ok, changed. h5py still needs to be built form source since there is no wheel for the mpi version afaik

bluerecording/__init__.py
cattabiani: I am not sure I got this comment. I changed the description to explain why

bluerecording/__init__.py
cattabiani: the idea is to use this setup.sh mostly locally. On the platform I expect a container of some sort with already the requirements installed. In that case you can just `pip install .` in this folder

setup.sh
cattabiani: to develop on this: Judit said that our stack is largely based on neuron 9.0.1 and she suggested to pin neurodamus et al. dependencies to this one

bluerecording/__init__.py
mgeplf: From a transparency point of view, it's better to run a script instead of sourcing it, especially when it's quite long - otherwise there are surprises that lurk within the code - for instance, normally I create my own virtualenv; if I source this, I lose whatever place I was in.  I find this quite non-standard.  In addition, getting people in the habit of "installing" software by `source`-ing something is not a good idea.

Maybe one could execute the script, and then print out then environment variables that need to be set?
Or, alternatively, create a file with just what needs to be set?

bluerecording/__init__.py
mgeplf: The problem is that it defeats the whole purpose of the `dependencies` of pyproject; to list the dependencies of the project.  Part of the whole transition to a non-executable pyproject.toml (as opposed to setup.py) was to make introspection of the dependencies possible, but this is purposefully circumventing it.

What happens if h5py is installed `the wrong way`?

bluerecording/__init__.py
mgeplf: the wording `may` means that it is known when it needs to be pinned, but it's not clear what these situations are: hence it would be helpful to point to where it's described what sort of pinning is done, which hopefully also explains why.

bluerecording/__init__.py
cattabiani: > What happens if h5py is installed the wrong way?

the code fails when you try to run it saying: h5py does not have mpi support. 

This message may be even ok. Down the line, we still to install the correct neurodamus-models. If you get the wrong version it is harder to understand that that was the problem

bluerecording/__init__.py
cattabiani: but you solution is way clunkier and error-prone

I will try to remove `source`... but I find it a downgrade. What if I devise a version that works also if you already have a venv set up? Do you have other practical cases where this setup.sh fails?

setup.sh
mgeplf: Like I said, historically this hasn't been the case.
We probably ran for for about 3 years on a dev version of neuron (since neuron 8 didn' have what was needed, so the difference between 8.2a (May 2022) and 9.0.0 (Oct 2025).

I don't know, however, what the future will bring - now that Goran isn't around to help with the release, I suspect new versions won't be released often, so the change you made for the `report.conf`, for instances, may take quite a while to be released as a version.

setup.sh
mgeplf: Perhaps something like `--delete-install`; `--clean-install`

bluerecording/__init__.py
mgeplf: > but you solution is way clunkier and error-prone

I disagree:
1) it makes it clear I don't need to rerun `setup.sh` each time; I know what variables need to be set (which is also useful if one is making a container or if one closes their shell, and want to re-setup the environment.)
2) I can chose to set them, rather than being hidden from me.

cattabiani: I guess it is a matter of personal taste. To me, I see much more probable that I will forget to set them. However, I see your point. I will find a way to get it both ways

setup.sh
cattabiani: historically in OBI we moved to pinning because we (Judit) got hurt by a misalignment of neuron/neurodamus that lead to a 3-day debugging session that could have been avoided by pinning. Anyway, I let @jplanasc  continue on this, I do not have strong opinions. I am just reporting her points

mgeplf: That's a separate concern - whether it was pinned to a version or a commit, it would have been the same issue (it was the wrong one); I'm saying that (likely) can't pin to a *version* because those aren't released frequently enough.

The three day debugging happened, iirc, before the 9.0.1 release (which was in Nov  2025), so you would have had to pin to a *commit* since no version had the required functionality.

cattabiani: the issue was the other way around. Neuron was updated and neurodamus was not yet. The new neuron version was not compatible with the old neurodamus version and the error was a little too generic. I was on holiday and I could not advise Judit

setup.sh
cattabiani: chanced to `--clean-install`. I think it is clearest of the options

bluerecording/__init__.py
cattabiani: ok, I updated the explanations. I let you check

setup.sh
mgeplf: The point still stands; at some point neurodamus needed to be in sync with a version of NEURON that wasn't released yet.

cattabiani: apart that I think still that it was the opposite: neuron was already at the latest version and the neurodamus released version was not release yet...

I do not get what you want? Not pin neuron and base everything on main?

WeinaJi: BlueRecording doesn't run simulation, NEURON is just required by Emodel files to load the morphology and create cell objects. I think the stable neuron version is sufficient for that task. In fact, We might need to be careful about the Neurodamus version, maybe pin a specific version?

mgeplf: the problem w/ a stable neuron version is that neurodamus may not support it; so one has to be able to specify *commits* for both of them, not releaesed version.

ie: 
```
NEURON_COMMIT=[....]
NEURODAMAUS_COMMIT=[...]
```
By calling it `_VERSION` it suggest it needs to be a released version, but that's not the case - there has to be the flexibility of using commits, without being misled that only versions are supported.

bluerecording/__init__.py
eleftherioszisis: Whatever dependencies the package has they have to be explicitly stated in the pyproject.toml. This is what is made for. This is the expectation of everyone reading a package.

 They are not soft dependencies, they are mandatory. Then you can make whatever checks during runtime for correct installation.

cattabiani: to me it is better to say that a dependency is missing than installing the wrong one. Anyway, I see that I am the minority, I will try to accommodate all the requirements

At any rate this does not solve all the problems. Take neuron:

- for a normal installation we can do pip install ... ok
- for a full installation we need to install from source and I cannot pass by pip. Pip is not aware (afaik) that neuron was installed. Thus, a subsequent `pip install bluerecording` re-installs the neuron wheel (the wrong one) even if neuron is already there. The only way to solve this problem is to make `neuron` optional in `pyproject`. Would it be ok for you @eleftherioszisis ? In neurodamus it is already like this

cattabiani: @eleftherioszisis in addition... what do I do for `neurodamus-models`? It is a neuron/nmodl/cmake project, not a python one but a dependency for bluerecording

setup.sh
cattabiani: I just moved from `*_COMMIT` to `*_VERSION`. May we have a final decision on this? I can write whatever you want there, I do not particularly mind. 

About neurodamus, yes, I need or something after the latest released version because I added a breaking change in neurodamus right after and I need it. 

@mgeplf would it be fine for you to call `*_COMMIT` and then put a version in there? I think git does not mind in any case

`*_COMMIT_OR_VERSION` looks a little too long

cattabiani: @jplanasc may you give your opinion here too? Because I am trying to guess your opinion on this based on the comments on my previous prs...

cattabiani:
after various discussions I go back to draft. I need to cook something that is accepted by everybody

bluerecording/__init__.py
cattabiani: @mgeplf I updated the script quite heavily. I hope you like this version more :D

setup.sh
jplanasc: I can see that either pinning versions or commits have pros and cons. As @WeinaJi says, I think for BlueRecording it's ok to pin the NEURON version.

My main take would be that if we can pin a version, we should pin a version, which I think it's easier to track and it should be more tested than a specific commit.

However, I understand that in Neurodamus we may want to pin specific commits, especially for testing compatibility, but I think here is not critical.

And as @cattabiani said, we have our cloud set ups working with the same NEURON version, so everything is consistent. My first comment was mainly triggered because BlueRecording was pinning a commit that was older than NEURON 9.0.1, when that version had already been largely tested on the cloud environments, so I thought it made sense to update the pinning.

Anyway, I don't want to impose my thinking, but I'd prefer that we pin versions instead of commits whenever this is possible.

bluerecording/__init__.py
jplanasc: If I may add here, I'd expect python dependencies to be listed in `pyproject.toml`, but I agree that, for example, `h5py` is a hard one, because the default installation is without MPI. Therefore, it needs to be installed separately with specific flags that you can't specify in the pyproject file.

cattabiani: yes, the problem is that one. I could list every dep in pyproject.toml but this risks to install the wrong versions if things are not already in place. I see that people are fine with this risk so I am listing everything I can there. Still, neurodamus-models will be missing (not a python project) and neuron is optional because when installing from source pip does not see it and installs it twice (without Mike's magic trick that I do not want to use)

setup.sh
mgeplf: > which I think it's easier to track and it should be more tested than a specific commit.

I disagree; the commit is the only thing that is stable, and you know it's what you expect.
I believe that NEURON, in the past, did the (in)famous thing of releasing the different source under the same name, at one point - so, beware.

> we pin versions instead of commits whenever this is possible

That's fine; however, it should be clear from the name of the variable that a commit is allowed; by naming it `VERSION`, we don't know what is allowed, and it's easy to think that it is being installed by something that only allows versions.
If it keeps it's original name, then it's clear that it can be changed to any commit (and, likely will, if we have to add any features/fixes to NEURON) - and we can use the `tag` value if it's important that it's human readable.

bluerecording/__init__.py
mgeplf: > [...] neuron is optional [...]

Since it's a required dependency, I would keep it in pyproject.
To prevent the installation problem, may it's worth doing `pip install --no-deps` to install something without it overwriting the installed neuron.

I think we can probably do a less hacky trick to install neuron, I just haven't had time to try it yet...stay tuned.

cattabiani: the problem is not the override, it is just a duplicate. I am not sure which one is picked then

I will stay tuned. May we proceed with this solution for now? I am happy to come back to this once your magic is ready. The dep is there and there is a comment to explain

setup.sh
jplanasc: I still think a NEURON version is enough for this case, but ok, let's try to converge on something.

So, if I'm reading correctly, I think that everyone will be happy if:
- The variable name is not changed (kept as `NEURON_COMMIT`)
- We use a tag / version to make it more human readable whenever this is possible

Is that correct?

mgeplf: I think that makes the most sense, yes.

setup.sh
```diff
@@ -127,143 +139,176 @@ elif [[ "$OS" == "Linux" ]]; then
...
-            -DNRN_ENABLE_INTERVIEWS=OFF \
-            -DNRN_ENABLE_CORENEURON=ON \
-            -DCMAKE_C_COMPILER=$NRN_C_COMPILER \
-            -DCMAKE_CXX_COMPILER=$NRN_CXX_COMPILER \
-            -DCORENRN_ENABLE_REPORTING=ON \
-            -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR
-
-        cmake --build nrn/build --parallel
-        cmake --build nrn/build --target install
+    if [[ "$OS" == "Darwin" ]]; then
```
mgeplf: is this needed since the outcome is the same? Or should Darwin be using clang?

.github/workflows/ci.yml
```diff
@@ -18,31 +21,68 @@ jobs:
...
+          HDF5_DIR: /usr/lib/x86_64-linux-gnu/hdf5/openmpi
+        run: |
+          pip install --upgrade pip setuptools wheel cython numpy
+          pip install mpi4py
+          pip install --no-cache-dir --no-binary=h5py h5py --no-build-isolation
+          pip install git+https://github.com/openbraininstitute/neurodamus.git@${{ env.NEURODAMUS_COMMIT }}
+          pip install -e ".[neuron,test]"
 
-      - name: Cache venv
+      - name: Cache neurodamus-models
```
mgeplf: I seem to recall the models compile quickly, is this cache needed?
What's the difference in speed if it's not used?

mgeplf (commented):
Looks reasonable to me, I think someone else should review it though.
Thanks for the `source` change.

setup.sh
cattabiani: I guess we found middle ground on this and the name is already there (nothing else to do). I close this

.github/workflows/ci.yml
cattabiani: it is probably small. I guess you prefer no cache. I will remove it. I thought, since it is a slow-moving repo, we can cahce that

setup.sh
cattabiani: you are right. It is legacy code that came from before and I did not notice that the 2 branches became the same. Fixed

bluerecording/__init__.py
```diff
@@ -3,11 +3,36 @@
...
+
+    if not h5py.get_config().mpi:
+        raise ImportError(
+            "h5py is installed but was built without MPI support.\n"
+            "bluerecording requires parallel HDF5 I/O.\n"
+            "Fix with:\n"
+            "  pip uninstall h5py\n"
+            "  HDF5_MPI=ON pip install --no-cache-dir --no-binary=h5py h5py "
+            "--no-build-isolation\n"
+            "Or use 'source setup.sh' which handles this automatically."
```
jplanasc: I guess now `source` is not specifically needed? I'd adapt the wording to `use the setup script` maybe?

setup.sh
```diff
@@ -1,41 +1,49 @@
...
     case $arg in
-        --dev) INSTALL_MODE="dev" ;;
         --no-system) SKIP_SYSTEM=1 ;;
         --data) DOWNLOAD_DATA=1 ;;
-        --no-cache) NO_CACHE=1 ;;
-        --quick) QUICK=1 ;;
+        --clean-install) CLEAN_INSTALL=1 ;;
         --help|-h)
-            echo "Usage: source setup.sh [OPTIONS]"
+            echo "Usage: source setup.sh [OPTIONS]  or  ./setup.sh [OPTIONS]"
```
jplanasc: I know this was requested to change, but I'm not convinced by this approach: I'd keep installation and env set up in separate scripts. I find it confusing that you can either `source` or execute this script.

bluerecording/__init__.py
cattabiani: ah yes. I always forget the readme and init comments

cattabiani: done

setup.sh
cattabiani: The separation already exists: `./setup.sh` builds, `source env.sh` activates if you want. On the other hand I prefer to automate as much as possible. This can be achieved with `source setup.sh`. I can get it both ways, perfect for everybody's preference. 

Splitting further just means two scripts that need to stay in sync and an extra step for every user. The dual-mode isn't something users need to think about: if they `./setup.sh` by habit instead of sourcing, it tells them to `source env.sh`. It's a safety net, not complexity. I'd rather not add friction to everyone's workflow over a style preference.

mgeplf: > I'd keep installation and env set up in separate scripts.

That's my preferred approach too; the `env setup` can be created by the installation script.
This follows the "normal" convention (ie: like `python -mvenv venv` for creation, and `source venv/bin/activate` to prepare shell) - I can't remember examples of sourcing a shell script that does anything but setting some environment variables.

jplanasc: > Splitting further just means two scripts that need to stay in sync

I'd disagree with this part, as if it's the setup script that creates the env, whenever you install, you should have a sync'ed env file.

cattabiani: ok, updated. Now there is a dev_setup.sh which produces env.sh. run_tests will notice that env.sh was not sourced and will complain about that

examples/single_cell_l5_tpc/README.md
```diff
@@ -4,7 +4,7 @@ Extracellular recordings from a single layer 5 thick-tufted pyramidal cell place
 
 ## Prerequisites
 
-Run `source setup.sh --dev --data` from the project root. This installs all dependencies and downloads the required datasets.
+Run `./dev_setup.sh --data` from the project root, then `source env.sh`. This installs all dependencies and downloads the required datasets.
```
jplanasc: I guess this sentence has gone through several modifications, to me, it would look clearer as something like:

> Run `./dev_setup.sh --data` from the project root. This installs all dependencies, downloads the required datasets and creates an environment script `env.sh`. Then `source env.sh` to set up the project environment.

examples/sscx_100_cells/README.md
```diff
@@ -4,7 +4,7 @@ Extracellular recordings from a network of 100 SSCx cells.
 
 ## Prerequisites
 
-Run `source setup.sh --dev --data` from the project root. This installs all dependencies and downloads the required datasets (networks and atlas).
+Run `./dev_setup.sh --data` from the project root, then `source env.sh`. This installs all dependencies and downloads the required datasets (networks and atlas).
```
jplanasc: Same comment as above.

README.md
```diff
@@ -10,70 +10,79 @@ This branch provides code that produces an electrodes file compatible with the [
...
+| `h5py` | Declared in pyproject.toml, but must be the MPI-enabled build (`HDF5_MPI=ON`). The default pip wheel lacks MPI support. |
+| `neuron` | Optional extra (`pip install bluerecording[neuron]`). For simulations with reporting, build from source instead. |
+| `neurodamus` | Declared in pyproject.toml. `dev_setup.sh` may install from a specific Git commit. |
+| `neurodamus-models` | CMake project (not a Python package). Required in all cases — needed to generate weights or run simulations. Built by `dev_setup.sh`. |
+
+If h5py lacks MPI support or neuron is missing, bluerecording will raise a clear error on import.
+
+The provided `dev_setup.sh` script handles all of this automatically, building NEURON and libsonatareport from source. It works on macOS (with `brew`) and Linux (with `apt`).
+
+**Development setup** — builds everything from source with test and notebook dependencies:
```
jplanasc: I'd reorganize the way you explain things in the readme:
- Move installation steps outside the `Dependencies` heading, I think they should be at the same level: first, you explain which dependencies you need to install, then, you explain how to install it. I don't think the steps should be nested inside the dependencies
- I'd first put the user installation steps and leave dev installation at the end

README.md
```diff
@@ -10,70 +10,79 @@ This branch provides code that produces an electrodes file compatible with the [
...
+| Package | Notes |
+|---|---|
+| `h5py` | Declared in pyproject.toml, but must be the MPI-enabled build (`HDF5_MPI=ON`). The default pip wheel lacks MPI support. |
+| `neuron` | Optional extra (`pip install bluerecording[neuron]`). For simulations with reporting, build from source instead. |
+| `neurodamus` | Declared in pyproject.toml. `dev_setup.sh` may install from a specific Git commit. |
+| `neurodamus-models` | CMake project (not a Python package). Required in all cases — needed to generate weights or run simulations. Built by `dev_setup.sh`. |
+
+If h5py lacks MPI support or neuron is missing, bluerecording will raise a clear error on import.
+
+The provided `dev_setup.sh` script handles all of this automatically, building NEURON and libsonatareport from source. It works on macOS (with `brew`) and Linux (with `apt`).
```
jplanasc: If this is only for developers, I'd remove it from this section and put it just in the developers instructions. I think it may be confusing here.

README.md
```diff
@@ -10,70 +10,79 @@ This branch provides code that produces an electrodes file compatible with the [
...
 ```bash
-source setup.sh --dev --no-cache
+pip install bluerecording
 ```
 
 Finally, if you want to run the full testing suite you need `--data`. See the [Testing](#testing) section for details.
 
-
-This is required only once to set up your python virtual environment. In future sessions you still need to run once:
+This is required only once to set up your python virtual environment. In future sessions you still need to activate the environment:
```
jplanasc: > This is required only once

What's `this`? I think it's not clear what you mean.

examples/single_cell_l5_tpc/README.md
cattabiani: done

examples/sscx_100_cells/README.md
cattabiani: done

README.md
cattabiani: done. I let you check the readme

README.md
cattabiani: it has been reorganized. Maybe it is already enough? I let you check

README.md
cattabiani: removed because it is outdated. I let you check

README.md
```diff
@@ -10,70 +10,75 @@ This branch provides code that produces an electrodes file compatible with the [
...
+Use `--clean-install` to wipe the virtual environment and all build artifacts before reinstalling from scratch (downloaded data is preserved):
 
 ```bash
-source setup.sh
+./dev_setup.sh --clean-install
+source env.sh
 ```
 
-But the execution is much faster. 
+If you want to run the full testing suite you need `--data`. See the [Testing](#testing) section for details.
```
jplanasc, line 56: I think this formatting is wrong? Probably missing an extra blank line below?

jplanasc (dismissed):
Just a minor formatting detail, the rest LGTM.

Good to go for me, as long as the other reviewers agree.

README.md
cattabiani: yes, true

jplanasc (dismissed):
Good to go from my side, as long as the other reviewers agree.

dev_setup.sh
```diff
@@ -0,0 +1,363 @@
...
+        -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR
+
+    cmake --build nrn/build --parallel
+    cmake --build nrn/build --target install
+fi
+
+# -------------------------
+# neurodamus
+# -------------------------
+# TODO: Switch back to the PyPI release once it includes commit e5abd33
```
mgeplf: this is no longer relavent, right?

dev_setup.sh
```diff
@@ -0,0 +1,363 @@
...
+        echo "Unsupported OS: $OS"
+        exit 1
+    fi
+fi
+
+# -------------------------
+# Set globals
+# -------------------------
+export NEURODAMUS_NEOCORTEX_ROOT="$(pwd)/neurodamus-models/build/install"
+export HOC_LIBRARY_PATH="$NEURODAMUS_NEOCORTEX_ROOT/share/neurodamus_neocortex/hoc"
```
mgeplf, line 125: is this ever used except to make the env file?

dev_setup.sh
cattabiani: yes, true. I leave the pip install from git as we may want to pin a commit

dev_setup.sh
cattabiani: yes, this one is there just to be set later. However, I prefer to have all the exports in one place

mgeplf (approved):

PR #60, opened 2026-04-24
update weights

### LINKED ISSUE #58: refactor and update `weights.py`
We should finally tackle `weights.py` and rework all its functions

cattabiani:
## Context

Systematic refactoring of `bluerecording/weights.py`, function by function. Based on `katta/update_morph`.

Fix: #19 #58 

## Scope

- **`ElectrodeFileStructure` class → `write_electrode_metadata_to_h5` function**: the class was a side-effect writer with dead state (`_ids` was never read). Replaced with a plain function.
- **`ElectrodeType` StrEnum**: replaces the `electrode_type()` validation function and all string comparisons with proper enum members.
- **`ObjectiveCSDParams` dataclass**: replaces the untyped dict that was used to pass objective CSD parameters around.
- **`Electrode` dataclass with `from_csv` factory**: replaces `make_electrode_dict`. Electrode metadata is now a typed dataclass instead of a nested dict.
- **CSV format change for ObjectiveCSD**: radius and thickness are now separate CSV columns instead of being encoded in the type string (`ObjectiveCSD_Disk_100_50`). `process_objectiveCSD` and `check_input_type_objectiveCSD` are removed.
- **`write_h5_file` split**: extracted `_compute_electrode_coeffs` (dispatch loop) and `_write_neurite_types` into separate functions. Dropped C-style return codes.
- **`write_all_neuron` → `_init_scaling_factors_and_offsets`**: renamed to reflect what it does, made private.
- **Single-rank H5 fallback**: `write_h5_file` uses plain `h5py.File` when running on a single MPI rank, avoiding the `mpio` driver requirement.

## Across all functions

- Type hints on all function signatures
- camelCase → snake_case for function names and local variables
- Improved docstrings with Args sections
- f-strings replacing string concatenation
- Removed excessive blank lines
- Narrowed bare `except:` to specific exception types

## Bug fixes

- **`get_coeffs_reciprocity`**: fixed in-place mutation of caller's DataFrame (`compartment_positions *= 1e-6` → work on a copy)
- **`line_source_cases`**: added guard for unhandled geometry case (previously returned unbound variable)
- **`electrode_type`**: was raising `AssertionError` (typo, not a real builtin) — now uses `ValueError` via the enum

## Linting and formatting

- Use tox for lint and format
- Added `ruff` as linter and formatter
- Configuration in `pyproject.toml` (line-length 120, select E/W/F/I/UP/B/SIM)
- Added `lint` job to `.github/workflows/ci.yml` (runs before tests)
- Fixed all lint issues across the codebase (import sorting, unused imports, bare excepts, `zip(strict=True)`, etc.)
- Applied `ruff format` to all source and test files

## Other

- Removed unused imports: `os`, `sys`, `json`, `datetime`, `scipy.spatial.distance`, `scipy.spatial.transform.Rotation`, `from .utils import *`
- `get_coeffs_objective_csd_sphere`: removed dead `get_array_spacing` call
- `calculate_axial_vectors`: replaced vstack loop with `np.tile`
- Updated `ObjectiveCSD.md` to document the new CSV column format
- Added `tests/data/electrode_objective.csv` test fixture
- All existing tests pass (28 unit + 4 integration)


bluerecording/weights.py
```diff
@@ -1,204 +1,191 @@
...
+        ``type`` (default ``LineSource``), ``layer``, ``region``,
+        ``radius``, ``thickness``.  The last two are only used for
+        ObjectiveCSD electrode types.
+        """
+        electrode_df = pd.read_csv(electrode_csv, header=0, index_col=0)
+
+        electrodes: dict[str, Electrode] = {}
+
+        for i in range(len(electrode_df)):
+            name = electrode_df.index.values[i]
```
WeinaJi: I see you need to store this name for writing the weights, how about just add `name` in the Electrode dataclass? It can be an attribute of the Electrode.

bluerecording/cli.py
```diff
@@ -1,86 +1,59 @@
...
-        help="Do not replace existing axons (default: replace)"
+        help="Do not replace existing axons (default: replace)",
     )
     ww_parser.add_argument(
         "--sigma",
         type=float,
         nargs="+",
         default=None,
-        help=f"Extracellular conductivity in S/m (default: {DEFAULT_SIGMA})"
+        help=f"Extracellular conductivity in S/m (default: {DEFAULT_SIGMA})",
```
WeinaJi: The help info mentions `[DEFAULT_SIGMA](default: {DEFAULT_SIGMA})`. Can you just set `default=[DEFAULT_SIGMA]` here ?

bluerecording/weights.py
```diff
@@ -256,638 +239,676 @@ def initialize_h5_file(cols, population_name, outputfile, electrode_csv, with_ne
...
-            coeffs = get_coeffs_lineSource(positions,columns,epos,sigma[sigmaIdx])
+    for electrode_idx, electrode in enumerate(electrode_names):
+        epos = h5["electrodes"][str(electrode)]["position"][:]
+        electrode_type = ElectrodeType(h5["electrodes"][str(electrode)]["type"][()].decode())
 
+        if electrode_type is ElectrodeType.LINE_SOURCE:
+            coeffs = get_coeffs_line_source(positions, columns, epos, sigma[sigma_idx])
             if len(sigma) > 1:
-                sigmaIdx += 1
+                sigma_idx += 1
```
WeinaJi, line 793: I don't understand the order in the `sigma` list. why doesn't it need to be incremented for "ObjectiveCSD*" types? `len(sigma)` can be different than the size of the `electrodes`?

bluerecording/weights.py
cattabiani: it would create a redundancy and I would need to keep things in sync. It is against single source of truth paradigm. I wouldn't do that

If you want to turn this in a list... yes, in that it makes sense. However, the tests become slightly more complex to write

WeinaJi: Yes, I want to turn it into a list, and store the name into the Electrode object.

bluerecording/cli.py
cattabiani: I'd keep default=None so that write_weights stays the single source of truth for the default sigma. Otherwise the default logic lives in two places (CLI and Python API) and they could drift apart. The None sentinel also makes it clear whether the user explicitly passed --sigma or not.

bluerecording/weights.py
cattabiani: sigma is only used by LINE_SOURCE and POINT_SOURCE electrodes — ObjectiveCSD* types don't depend on conductivity, so sigma_idx isn't incremented for them. If len(sigma) == 1 it's a uniform conductivity for all; if > 1, each line/point source electrode gets its own value in order. This was the original logic and I didn't want to change it in this PR — it may be reviewed in the future

bluerecording/weights.py
cattabiani: A list would be more natural if CSV order was the canonical order (index → matrix column directly). But since we sort alphabetically anyway to guarantee deterministic ordering, a list doesn't simplify anything — you still need the sorting step. And the dict gives you O(1) lookup by name for free, which is useful in tests and potential post-processing. No reason to switch IMHO

bluerecording/cli.py
WeinaJi: Then there is no need mention the (default: {DEFAULT_SIGMA})" in the help msg. The default is None for this argument via CLI.

bluerecording/weights.py
WeinaJi: Where do you sort alphabetically? 

> And the dict gives you O(1) lookup by name for free

Where do you need to look up by name? I see only the iteration in the original order of the dict in function `write_electrode_metadata_to_h5`. That's why I think a list is just sufficient. 
```
for index, (key, electrode) in enumerate(electrodes.items()):
```

bluerecording/weights.py
WeinaJi: When sigma is a list, the first few elements are always for line/point source electrodes? How can we make sure of order? Yes, it is better to review it in the future. For this PR, it is ok to keep it.

cattabiani: issue added. I agree that this should change:

https://github.com/openbraininstitute/BlueRecording/issues/73

bluerecording/cli.py
cattabiani: I do not agree. the `DEFAULT_SIGMA` is substituted with the default value (which is what we are going to use in the end). Otherwise users may ask: what is the default value? 

I know that it requires knowing that DEFAULT_SIGMA will become the default and we need to keep it in sync (be sure that in case of None we actually use this) but I thought that was not changing really in the future (apart from a numerical change of the variable which would remain synced automatically)

Screenshot 2026-05-08 at 09 24 01

cattabiani: Just to make sure we're on the same page, did you catch that:

```
from .weights import DEFAULT_SIGMA, initialize_h5_file, write_h5_file
```

bluerecording/weights.py
cattabiani: the sorting happens here:

```
electrode_names = sort_electrode_names(h5["electrodes"].keys(), population_name)
```

the access by name is used, for now, only in the tests. Given the number of electrodes in the tests it does not matter how we store them. 

I suspect that in production code it may be useful and there may be many electrodes... however, atm I do not know. 

I think the name sorting is because HDF5 does not guarantee key iteration order. We have already an order enforced by the index here:

```
h5.create_dataset(f"{prefix}/{population_name}", data=index))
```

maybe we could use that instead. 

Do you prefer a list and sorting based on this (or name)? I will not die on this hill. If you insist I can change it as you want

bluerecording/cli.py
WeinaJi: I mean the `DEFAULT_SIGMA` is used in `weights.py` not here. Here `args.sigma` is None if not specified.
Maybe you can modify the help msg "Extracellular conductivity in S/m (use {DEFAULT_SIGMA} if not specified)" to avoid misunderstanding. Later if you update the default behaviour in `weights.py`, remember to update this help msg as well.

bluerecording/weights.py
WeinaJi:
```
electrode_names = sort_electrode_names(h5["electrodes"].keys(), population_name)
```
I am confused by the sorting and access by name, why is it needed? It doesn't sort the datasets already written in the output h5 file. Is that to align with the `sigma` list for computing coefffs?  Anyway, atm we can keep the sorting by name. 

> Do you prefer a list and sorting based on this (or name)? I will not die on this hill. If you insist I can change it as you want

Yes, I prefer a list, and to turn the name into the attribute. In `write_electrode_metadata_to_h5`, then it just changes to `prefix = f"electrodes/electrode.name"`. The rest will be the same I think.

cattabiani: ok, now we have a list. I let you check

bluerecording/cli.py
cattabiani: ok, done

WeinaJi (dismissed):
LGTM. 
@jamesgking , you mentioned that you ran the local tests for this PR. Does everything pass ?

jamesgking:
I confirm all tests pass on my local machine

WeinaJi (approved):
LGTM!

PR #65, opened 2026-04-28
add PyPI publish workflow on tag push

### LINKED ISSUE #63: release to pypi
To have BlueRecording incorporated in the normal OBI tasks we need to do a release on pypi

cattabiani:
## Context

Fix #63

## Scope

Adds a GitHub Actions workflow (`.github/workflows/publish.yml`) that:

- Builds sdist + wheel on every push, PR, and manual trigger (catches build issues early)
- Publishes to PyPI automatically when a tag is pushed
- Uses OIDC trusted publishing (no API tokens needed)
- Follows the same pattern as neurodamus's `publish_wheels.yml`

## Before merging

A Trusted Publisher must be registered on PyPI:
- PyPI project name: `bluerecording`
- Owner: `openbraininstitute`
- Repository: `BlueRecording`
- Workflow name: `publish.yml`

Once merged, pushing a semver tag (e.g. `0.4.0`) will trigger the publish.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,54 @@
+name: Publish to PyPI
+
+on: [pull_request, push, workflow_dispatch]
+
+jobs:
+  wheel:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python_version: ["3.11"]
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python_version }}
+
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # needed for hatch-vcs to resolve version from tags
+
+      - name: Build Wheel
+        run: |
+          cd $GITHUB_WORKSPACE
+          python${{ matrix.python_version }} -m pip install --upgrade build
+          python${{ matrix.python_version }} -m build .
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: release-artifacts-${{ matrix.python_version }}-${{ strategy.job-index }}
+          path: ./dist/*
+
+  pypi-publish:
+    runs-on: ubuntu-22.04
+    needs: [wheel]
+    permissions:
+      id-token: write
+    steps:
+      - name: Download artifacts produced by the wheel job
+        uses: actions/download-artifact@v4
+        with:
+          pattern: release-artifacts*
+          path: dist/
+          merge-multiple: true
+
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+      matrix:
+        python_version: ["3.11"]
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python_version }}
+
+      - name: Checkout Repository
+        uses: actions/checkout@v4
```
heerener: Same here: v6 is also latest & greatest.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  wheel:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python_version: ["3.11"]
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v5
```
heerener: I recommend v6; v5 is going to cause deprecation warnings.

See e.g. https://github.com/openbraininstitute/neurodamus/actions/runs/24998871539

.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          fetch-depth: 0  # needed for hatch-vcs to resolve version from tags
+
+      - name: Build Wheel
+        run: |
+          cd $GITHUB_WORKSPACE
+          python${{ matrix.python_version }} -m pip install --upgrade build
+          python${{ matrix.python_version }} -m build .
+
+      - uses: actions/upload-artifact@v4
```
heerener: This one will also generate deprecation warnings; latest is v7

.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+          path: ./dist/*
+
+  pypi-publish:
+    runs-on: ubuntu-22.04
+    needs: [wheel]
+    permissions:
+      id-token: write
+    steps:
+      - name: Download artifacts produced by the wheel job
+        uses: actions/download-artifact@v4
```
heerener: v8 here

.github/workflows/publish.yml
cattabiani: done

.github/workflows/publish.yml
cattabiani: done

.github/workflows/publish.yml
cattabiani: done

heerener (approved):

Issue #59, opened 2026-04-24
make this notebook work


cattabiani:
@james-isbister pointed at this [notebook](https://github.com/openbraininstitute/obi-one/blob/extracellular_electrode_array/examples/J_create_recording_array/create_recording_array.ipynb) as the benchmark on which the hooks from obi-one will be shaped to use `BlueRecording`

evakennyobi:
There's an update here @cattabiani ? thanks

cattabiani:
yes. It works. Can we close this @james-isbister ?

james-isbister:
Yes, thanks @cattabiani

PR #67, opened 2026-04-29
Split data download into standalone script, expand mpi4py comment, drop bytewise test

### LINKED ISSUE #62: `NodePopulation.get_attribute` expects a raw gid
[this](https://github.com/openbraininstitute/BlueRecording/blob/46bd0fdb868f18352be03901f49e2e1822e28ae7/bluerecording/positions.py#L571) is a bug. Currently it is not picked up because the pop_offsets are 0 for all the tests that we have. We should use `cell.raw_gid`

### LINKED ISSUE #66: BlueRecording tests fail with seg. fault in the docker image
[This image](https://github.com/openbraininstitute/launch-system/pull/148/changes#diff-42095dd72960fa9d60ba9cd5ace5bef9d6188b5719d14254af72d4b0d7e5338c) should run `BlueRecording` for the platform. In particular, it provides an env with neuron (with libsonatareport), h5py with mpi, neurodamus and neuron to fully run `BlueRecording`. 

@eleftherioszisis yesteraday tried to run BlueRecording tests. The unit tests worked but the more serious ones failed with seg. fault.

We (I) need to investigate what is going on over there

cattabiani:
## Context

Fix: #66 #62

## Scope

- Extract data download from dev_setup.sh into download_examples_data.sh (installs unzip on Linux if missing, idempotent)
- Remove --data flag from dev_setup.sh; print hint for new script
- Expand mpi4py import comment re NRN_ENABLE_MPI_DYNAMIC=OFF
- Remove bytewise H5 comparison in test_circuit_write_weights (assert_allclose already covers numerical correctness)
- Update READMEs and run_tests.sh references
- use raw_gid for `get_attribute("morphology"...` (fix #62 bug. The test will come in a future PR, already in the pipeline)

### DIFF
--- a/bluerecording/__init__.py
+++ b/bluerecording/__init__.py
@@ -27,10 +27,10 @@ def _check_dependencies():
         )
 
     try:
-        # noqa: F401 — mpi4py must be imported before neuron to ensure MPI is
-        # initialized. This is necessary when NEURON is compiled with
-        # -DNRN_ENABLE_MPI_DYNAMIC=OFF, as NEURON will not dynamically load
-        # MPI support at runtime and expects MPI to already be initialized.
+        # noqa: F401 — mpi4py must be imported before neuron to ensure
+        # MPI_Init has been called. When NEURON is statically linked against
+        # MPI (NRN_ENABLE_MPI_DYNAMIC=OFF), it expects MPI to already be
+        # initialized by the time it loads.
         from mpi4py import MPI  # noqa: F401
         import neuron  # noqa: F401
     except ImportError:
--- a/bluerecording/__init__.py
+++ b/bluerecording/__init__.py
@@ -31,7 +31,7 @@ def _check_dependencies():
         # initialized. This is necessary when NEURON is compiled with
         # -DNRN_ENABLE_MPI_DYNAMIC=OFF, as NEURON will not dynamically load
         # MPI support at runtime and expects MPI to already be initialized.
-        import mpi4py  # noqa: F401
+        from mpi4py import MPI  # noqa: F401
         import neuron  # noqa: F401
     except ImportError:
         raise ImportError(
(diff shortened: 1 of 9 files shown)


bluerecording/__init__.py
```diff
@@ -27,6 +27,11 @@ def _check_dependencies():
         )
 
     try:
+        # noqa: F401 — mpi4py must be imported before neuron to ensure MPI is
+        # initialized. This is necessary when NEURON is compiled with
+        # -DNRN_ENABLE_MPI_DYNAMIC=OFF, as NEURON will not dynamically load
```
mgeplf: I don't think this is the case; `NRN_ENABLE_MPI_DYNAMIC`; this is for whether the MPI library will be loaded dynamically (so it doesn't have to exist on the target, but it'll be used if it does)

bluerecording/positions.py
```diff
@@ -566,7 +566,7 @@ def get_positions(node_manager, ids, cols, population, morphologies_dir, replace
     neurite_type_arrays = []
     for i in ids:
         cell = node_manager.get_cell(i)
-        m, center = get_morphology(population, i, morphologies_dir, cell)
+        m, center = get_morphology(population, cell.raw_gid, morphologies_dir, cell)
```
mgeplf, line 569: is this what was causing the issue w/ loading the morphology?

cattabiani: this was the problem that I discovered with the new circuit. The one that comes from this:

https://github.com/openbraininstitute/BlueRecording/issues/59

The problem was there since the beginning or I introduced it a long time ago and never noticed because the tests do not cover the `offset =/= 0` case

bluerecording/__init__.py
cattabiani: ok, let me rephrase it

mgeplf (approved):

Issue #53, opened 2026-04-21
create completely syntetic test (like ringtest) to expand testing capabilities


cattabiani:
As @WeinaJi pointed out our testing is still lacking. The recently merged fix #47 testifies that some parts of the code are untested. It would be better to have a ringtest-like test.

### Requirements:

- lightweight: it should be usable in the CI
- complex enough to catch various paths in the code

cattabiani:
we can use the new circuit `rat_s1_forelimb_l56_10cells`. I can catch 2 birds with 1 stone

evakennyobi:
This is in code review @cattabiani ? thanks

cattabiani:
maybe I can close this today. Weina's review pending

PR #70, opened 2026-05-04
circuit config support, new rat test, public API rework

### LINKED ISSUE #71: divide unit and integration tests
There are (not so) secret integration tests lurking in our unit tests. We should divide them

cattabiani:
## Context

This is all preliminary work for #59 

Meanwhile:

Fix: #71 #69 #53

## Scope

The CLI now accepts a `circuit_config.json` in addition to a `simulation_config.json`. When a circuit config is detected, a minimal temporary simulation config is generated on the fly and passed to neurodamus. This is fully transparent — existing simulation config workflows are unchanged.

A new 10-cell test circuit (`rat_s1_forelimb_l56_10cells`) is included as a lightweight fixture that runs in CI without needing external data downloads.

### Feature: circuit config support

- `bluerecording/utils.py`: added `is_circuit_config()`, `_make_simulation_config()`, and `resolve_config()` context manager. Detection is based on the presence of a top-level `"networks"` key (circuit config) vs `"network"` + `"run"` (simulation config).
- `resolve_config` is MPI-safe: rank 0 creates the temp file, broadcasts the path, and all ranks synchronize via barriers before and after use.
- `bluerecording/cli.py`: both `write_positions` and `write_weights` now wrap `init_circuit` inside `resolve_config`, accepting either config type. Help text updated accordingly.

### Test reorganization

- Moved integration tests (positions, weights) from `tests/unit/` and `tests/unit-mpi/` into `tests/integration/` and `tests/integration-mpi/`.
- `tests/unit/` and `tests/unit-mpi/` now contain only pure unit tests (morphology, interpolation, coefficients, H5 structure, parallel H5 I/O).
- Renamed test functions and files to use the circuit name (`sscx_100_cells`, `single_cell`, `rat_s1`) instead of generic `circuit`.

### New test circuit

- `examples/rat_s1_forelimb_l56_10cells/`: 10-cell L5/L6 rat S1 forelimb circuit with morphologies, node/edge h5 files, node sets, emodel hoc files, mod files, and an electrode CSV. ~24 MB committed directly (comparable to existing `single_cell_l5_tpc`).

### New tests (run in CI)

- `test_rat_s1_get_positions` — non-MPI smoke test via `resolve_config`
- `test_rat_s1_write_weights` — non-MPI write_weights pipeline
- `test_rat_s1_neurite_types` — neurite type verification (switched from single_cell to rat_s1 to enable in CI)
- `test_rat_s1_get_positions_mpi` — MPI positions with 2 ranks
- `test_rat_s1_write_weights_mpi` — MPI write_weights with 2 ranks

### CI and run_tests.sh

- `ci.yml`: added `tests/integration` step with `--forked`, MPI integration tests run one file per `mpirun` to avoid NEURON GID collisions.
- `run_tests.sh`: refactored to eliminate duplication — `ci` mode reuses `all` with a marker filter (`-m "not skip_in_ci"`). Added `integration` option. MPI integration tests use a glob instead of a hardcoded file list. Pre-filters files with no matching tests to avoid pytest exit code 5 under `mpirun`.
- `README.md`: updated Testing section to reflect new directory structure and commands.
- `.gitignore`: added `.bluerecording_sim_*.json` for temp file safety.


bluerecording/__init__.py
```diff
@@ -43,3 +43,9 @@ def _check_dependencies():
 
 
 _check_dependencies()
+
+# Public API
```
mgeplf: `# Public API` doesn't need to be a comment, it's implicit from the fact that it's being loaded in `__init__.py` and the names don't start with `_`

bluerecording/circuit.py
```diff
@@ -3,20 +3,24 @@
 
 Provides the entry point for loading a circuit model and extracting
 the discretization info (node IDs, compartment structure, morphology access)
-needed by both get_positions and write_weights.
+needed by both compute_positions and compute_weights.
```
mgeplf: I find that comments like `needed by both compute_positions and compute_weights.` go stale quickly and don't really provide actionable information, so I would remove it instead of updating it.

bluerecording/circuit.py
```diff
@@ -33,33 +37,34 @@ def init_circuit(path_to_simconfig: str):
...
-    ).reshape(-1, 2)
+        ids = node_manager.get_final_gids()
+        points = node_manager.target_manager.get_target(None).get_point_list(
+            node_manager,
+            libsonata.SimulationConfig.Report.Sections.all,
+            libsonata.SimulationConfig.Report.Compartments.all,
+        )
+        cols = np.array(
+            [(p.gid, s) for p in points for s in sorted(p.sclst_ids)],
+            dtype=np.int64,
```
mgeplf, line 57: why is this `np.int64`? can `sclst_ids` have negative ids?

bluerecording/cli.py
```diff
@@ -1,9 +1,10 @@
 import argparse
-from pathlib import Path
 
 from . import __version__, positions
 from .circuit import init_circuit
-from .weights import DEFAULT_SIGMA, initialize_h5_file, write_h5_file
+from .positions import compute_positions, save_positions
+from .utils import resolve_output_path
+from .weights import DEFAULT_SIGMA, Electrode, _get_weights, save_weights
```
mgeplf: I'm surprised linting isn't complaining about `_get_weights`, which appears to be private to the `.weights` module - maybe that needs to be checked as to why.

Is there a reason that `_get_weights` is "marked" private but is being used here?

bluerecording/cli.py
```diff
@@ -73,45 +74,38 @@ def main():
...
 
     elif args.command == "write_weights":
         node_manager, ids, cols, population, population_name, morphologies_dir = init_circuit(args.path_to_simconfig)
-        positions_df, cols, neurite_types = positions.get_positions(
+
+        output_file = resolve_output_path(args.output_path)
+
+        electrodes = Electrode.from_csv(args.electrode_csv)
+
+        positions_df, cols, neurite_types = positions._get_positions(
```
mgeplf: same question about `_get_positions` and why it's private if it's being used "publically"

bluerecording/utils.py
```diff
@@ -18,3 +26,119 @@ def get_circuit_path(path_to_simconfig: str | Path) -> str:
     """
     sim_conf = libsonata.SimulationConfig.from_file(str(path_to_simconfig))
     return sim_conf.network
+
+
+def is_circuit_config(path: str | Path) -> bool:
```
mgeplf: this only seems to be used in this module, should it be private?

bluerecording/utils.py
```diff
@@ -18,3 +26,119 @@ def get_circuit_path(path_to_simconfig: str | Path) -> str:
...
+        "run": {
+            "dt": 0.025,
+            "tstop": 0,
+            "random_seed": 1,
+        },
+    }
+
+
+@contextmanager
+def resolve_config(path: str | Path):
```
mgeplf: I think this should have a more clear name, perhaps something like `path_to_simulation_config` or `resolve_simulation_config`; resolve is quite a generic term, especially in a module named `utils`; it doesn't really describe what it does.

bluerecording/utils.py
```diff
@@ -18,3 +26,119 @@ def get_circuit_path(path_to_simconfig: str | Path) -> str:
...
+    if not path.is_file():
+        raise FileNotFoundError(f"Config file not found: {path}")
+
+    if not is_circuit_config(path):
+        yield str(path)
+    else:
+        tmp_name = None
+        if rank == 0:
+            sim_cfg = _make_simulation_config(path)
+            tmp = tempfile.NamedTemporaryFile(  # noqa: SIM115  # noqa: SIM115
```
mgeplf: why is this here twice `# noqa: SIM115  # noqa: SIM115`

bluerecording/weights.py
```diff
@@ -10,6 +10,9 @@
 from scipy.interpolate import RegularGridInterpolator
 from sklearn.decomposition import PCA
 
+from . import positions as _pos_module
```
mgeplf: perhaps `from . import positions as _positions`?

mgeplf (commented):
I had to stop reviewing, this PR should be 3 separate PRs:
1) adding the tests
2) changing the public API
3) config file support

Individually, these are easy reviews, but it takes me more than 3 times as long to do them when they are all grouped together, plus it's so big that it can't be done in 1 go.

In the future, please have the PR contain related changes.

cattabiani:
@mgeplf 

> I had to stop reviewing, this PR should be 3 separate PRs:
> 
>     adding the tests
>     changing the public API
>     config file support
> 
> Individually, these are easy reviews, but it takes me more than 3 times as long to do them when they are all grouped together, plus it's so big that it can't be done in 1 go.
> 
> In the future, please have the PR contain related changes.

I know it is not ideal. Conceptually this is all that is needed to make the notebook work:

- I need tests and the notebook provides a new one
- The tests are unsorted and all together in unit tests. Better splitting may simplify your review

I started the config file support on top of main first. Then I discovered that I was re-doing changes that were in #60 already. So I tried to rebase but there were too many commits already and too many conflicts. I had to squash and rebase

This is not an excuse. It is an explanation on how I reached this point

bluerecording/__init__.py
cattabiani: ok. removed

bluerecording/circuit.py
cattabiani: removed

bluerecording/circuit.py
cattabiani: mostly for interoperability. It can indeed be uint. However:
- pandas comparisons usually compare signed values
- The downstream code does things like cols[:, 0] == gid where gid comes from various sources. In this way we avoid subtle missmatch bugs

I can change it to uint if you insist. If you refer to the 64 part it is because it is a gid and they are 64 bits in neuron

bluerecording/cli.py
cattabiani: they come from a previous iteration and I forgot to remove the `_`. I will remove it for this and `_get_positions`

bluerecording/cli.py
cattabiani: same answer as before

bluerecording/utils.py
cattabiani: yes, private now

bluerecording/utils.py
cattabiani: it will go away soon with the new libsonata but yes, the name is too generic. I let you check the new solution

bluerecording/utils.py
cattabiani: lol. copy-paste + a small window. I did not see it

cattabiani: (removed 1. we need one because it needs to persist outside)

bluerecording/weights.py
cattabiani: yes, done

bluerecording/circuit.py
mgeplf: > pandas comparisons usually compare signed values

How will that impact the comparison?

>  In this way we avoid subtle missmatch bugs

What sort of mismatch bugs are you talking about? Can you give an example?

bluerecording/utils.py
```diff
@@ -18,3 +26,119 @@ def get_circuit_path(path_to_simconfig: str | Path) -> str:
...
+
+        try:
+            yield tmp_name
+        finally:
+            comm.Barrier()
+            if rank == 0:
+                Path(tmp_name).unlink(missing_ok=True)
+
+
+def resolve_output_path(path: str | Path, default_filename: str = "weights.h5") -> Path:
```
mgeplf: I think this can be simplified by not bothering with the `default`; just require a full path be provided.

examples/rat_s1_forelimb_l56_10cells/emodels_hoc/cADpyr_L5TPC.hoc
```diff
@@ -0,0 +1,486 @@
+/*
```
mgeplf, line 1: do different hoc files have an impact on the weight generation?
if not, it's simpler to not require them.

bluerecording/circuit.py
cattabiani: nothing that is really a problem. I just thought that since we need to compare with signed values anyway, it was not worth the effort of using uint

examples/rat_s1_forelimb_l56_10cells/circuit_config.json
```diff
@@ -0,0 +1,91 @@
+{
+  "components": {
+    "biophysical_neuron_models_dir": "",
+    "mechanisms_dir": "",
+    "morphologies_dir": "",
+    "point_neuron_models_dir": "",
+    "provenance": {
+      "id_mapping": "id_mapping.json"
+    },
+    "synaptic_models_dir": "",
```
mgeplf: where is `synaptic_models_dir` coming from? it's not something that's described in the sonata-extension.

Also, are any of the components needed?

examples/rat_s1_forelimb_l56_10cells/run_coordinate_instance.json
```diff
@@ -0,0 +1,34 @@
+{
```
mgeplf: what is this used for?

tests/integration-mpi/test_rat_s1_write_weights.py
```diff
@@ -0,0 +1,42 @@
...
+import numpy as np
+import pytest
+from mpi4py import MPI
+
+from bluerecording import positions
+from bluerecording.circuit import init_circuit
+from bluerecording.weights import Electrode, get_weights, save_weights
+from tests.conftest import EXAMPLE_RAT_S1
+
+comm = MPI.COMM_WORLD
```
mgeplf, line 12: do these need to be at the module scope (applies to all mpi tests)

tests/integration/test_positions.py
```diff
@@ -0,0 +1,72 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import pandas as pd
+import pytest
+
+from bluerecording import positions
+from bluerecording.circuit import init_circuit
+
+
+@pytest.mark.skip_in_ci
```
mgeplf, line 9: what determines if something is `skip_in_ci`?
Is it the existence of something under `examples/`?
If so why not use the existence of whichever files are required to mark if a test should be run?

bluerecording/utils.py
cattabiani: I'd keep the default. It lets the CLI just pass a directory without caring about the filename, and avoids duplicating that logic in every caller. The function is tiny and the heuristic is obvious, so it's not hiding complexity.

If you insist, I can remove it. I will not die on this hill

examples/rat_s1_forelimb_l56_10cells/emodels_hoc/cADpyr_L5TPC.hoc
cattabiani: I just copy-pasted the test since it is the proposed one from James and, by chance, it was the perfect size for the ci tests. 

Let me see if we can simplify this stuff a little

cattabiani: hoc files are just 0.1% of the total. I think we can leave them as they are. I have a small bias towards leaving the test as I got it so users are not confused when they see a test that is similar to what I got ... but a little simplified

examples/rat_s1_forelimb_l56_10cells/circuit_config.json
cattabiani: no, I did not bother to inspect the test closely since the morphologies are anyway more than 50% of the total and they cannot be removed (in principle)

anyway, removed the components block

examples/rat_s1_forelimb_l56_10cells/emodels_hoc/cADpyr_L5TPC.hoc
mgeplf: ok, works for me

examples/rat_s1_forelimb_l56_10cells/run_coordinate_instance.json
cattabiani: just to trace provenance. I think I can just put a readme and remove id_mapping and this

tests/integration/test_positions.py
cattabiani: good question. Basically skip_in_ci are all the tests that are too long (in theory. Atm I do not think there are any really deemed too long) or that require downloading gbs of data from zenodo.

In particular:

- simple_cell has 2 FEM fields for write_weights
- sscx_100_cells requires parts of the atlas package

You can find what is downloaded here:

https://github.com/openbraininstitute/BlueRecording/blob/katta/circuit-config-support/download_examples_data.sh

Here we have a circuit that is only 24Mb. I decided that it was small enough to keep it directly in the source code. 

I can also add it to a publication of bluerecording itself so we download from github but it is not polluting the code base. It is possible, just a little bit more work

tests/integration-mpi/test_rat_s1_write_weights.py
cattabiani: it is mostly for the sake of simplicity. Otherwise I need to write this in all the tests. Also I have seen it everywhere, I thought it was the convention. 

The only difference would be to test the mpi init inside __init__ but even in that case I am not sure that that would be tested correctly in the pytests. 

Why is this a problem?

tests/integration/test_positions.py
mgeplf: maybe it would make sense to split them into categories based off which datasets are needed, and test for and run when those datasets exist?

Anyways, that can be done in another PR.

cattabiani: good point. Issue created:

https://github.com/openbraininstitute/BlueRecording/issues/72

bluerecording/utils.py
mgeplf: >  It lets the CLI just pass a directory without caring about the filename, and avoids duplicating that logic in every caller. 

If only one file is going to be output, specifying the exact path to it ends up being safer and more ergonomic.  In my experience, this has bit us before - people don't read the directions, and find out they overwrote a file by accident, for instance.  Best to be explicit.

tests/integration-mpi/test_rat_s1_write_weights.py
mgeplf: I like to keep the namespace clean, but I guess we can keep it that way if it's convention.

bluerecording/utils.py
cattabiani: done

mgeplf: thanks, I think this is safer.

mgeplf (dismissed):
I'm fine with it, might be worthwhile to have someone else review.

bluerecording/circuit.py
```diff
@@ -33,33 +36,34 @@ def init_circuit(path_to_simconfig: str):
...
-    node_manager = next(iter(nd.circuits.node_managers.values()))
+    with resolve_simulation_config(path_to_config) as path_to_simconfig:
+        nd = neurodamus.Neurodamus(
+            path_to_simconfig,
+            disable_reports=True,
+            direct_mode=True,
+            build_model=True,
+            enable_coord_mapping=True,
+            keep_build=False,
+            simulator="NEURON",
```
WeinaJi: As simulator is "NEURON", there is no need to set `direct_mode`, `keep_build`, and `build_model` explicitly here.

bluerecording/utils.py
```diff
@@ -18,3 +26,89 @@ def get_circuit_path(path_to_simconfig: str | Path) -> str:
...
+            )
+            json.dump(sim_cfg, tmp, indent=2)
+            tmp.close()
+            tmp_name = tmp.name
+
+        tmp_name = comm.bcast(tmp_name, root=0)
+        comm.Barrier()
+
+        try:
+            yield tmp_name
```
WeinaJi, line 110: Why does the `try` not include the previous steps that creates the `tmp` file? If  an exception is raised earlier, it won't be caught.

bluerecording/weights.py
```diff
@@ -734,21 +735,27 @@ def _parse_index_range(spec: str) -> range:
...
+    Args:
+        electrode_type: The ObjectiveCSD electrode type to match.
+        objective_csd_array_indices: Optional list of 'start:end' range specs.
+        objective_csd_count: Running count of CSD arrays encountered so far.
+        electrodes_ordered: List of Electrode objects in sorted order.
+        electrode_idx: Index of the current electrode in the sorted list.
     """
     if objective_csd_array_indices is None:
-        all_types = [h5["electrodes"][str(e)]["type"][()].decode() for e in electrode_names]
+        all_types = [e.type.type if isinstance(e.type, ObjectiveCSDParams) else e.type for e in electrodes_ordered]
```
WeinaJi: `e.type.type` ? Is there a better naming for the 2nd type ?

.github/workflows/ci.yml
```diff
@@ -85,11 +85,37 @@ jobs:
           export PATH="${{ env.NEURODAMUS_NEOCORTEX_ROOT }}/bin:$PATH"
           pytest -v -m "not skip_in_ci" tests/unit
 
-      - name: Run MPI tests
+      - name: Run integration tests
+        env:
+          HOC_LIBRARY_PATH: ${{ env.NEURODAMUS_NEOCORTEX_ROOT }}/share/neurodamus_neocortex/hoc
+          CORENEURONLIB: ${{ env.NEURODAMUS_NEOCORTEX_ROOT }}/lib/libcorenrnmech.so
+          NRNMECH_LIB_PATH: ${{ env.NEURODAMUS_NEOCORTEX_ROOT }}/lib/libnrnmech.so
```
WeinaJi: If the values of these env vars are the same for all the tests, you can consider to makes the environment variable persist for all subsequent steps in the GitHub Actions workflow, like
```
        echo "HOC_LIBRARY_PATH=$HOC_LIBRARY_PATH" >> $GITHUB_ENV;
        echo "CORENEURONLIB=$CORENEURONLIB" >> $GITHUB_ENV;
        echo "NRNMECH_LIB_PATH=$NRNMECH_LIB_PATH" >> $GITHUB_ENV;
```

WeinaJi (commented):
Just some nitpicks

bluerecording/circuit.py
cattabiani: true. I added them before adding neuron and then I did not remove it .... Done

bluerecording/utils.py
cattabiani: ok

bluerecording/weights.py
cattabiani: true. let's be explicit

.github/workflows/ci.yml
cattabiani: true, I did not notice it became not DRY. done

WeinaJi (approved):
LGTM!

PR #82, opened 2026-06-18
Vectorize line-source weight computation (2)

### LINKED ISSUE #79: speed up: line source computation can be vectorized with numpy

cattabiani:
Refactor physics module: separate concerns, unify APIs, clean up units.

Fix: https://github.com/openbraininstitute/BlueRecording/issues/79

## Changes

- Extract physics tests into `tests/unit/test_physics.py`
- Extract `physics.py` module from `weights.py` with clear separation:
  - `SegmentGeometry.from_positions()`: parse geometry once, store in µm
  - `_line_source_coeffs`: vectorized kernel (µm + S/m → mV/nA)
  - `_point_source_coeffs`: vectorized kernel (µm + S/m → mV/nA)
  - `get_coeffs_line_source`: accepts single `(3,)` or multi `(N_elec, 3)` electrode positions
  - `get_coeffs_point_source`: same pattern
- No separate `_batch` functions — single/multi handled by one function via broadcasting
- Memory chunking lives in `get_weights()` via `electrode_chunk_size` parameter (exposed in CLI)
- `get_coeffs_dipole_reciprocity` computes center internally
- Add `log_rank0` helper to `utils.py`, use consistently
- Remove dead code: `precompute_segment_geometry` wrapper, `_get_line_coeffs`, `_line_source_cases`
- Remove redundant comments that restate the code
- Rename `testget_*` → `test_get_*` for consistency
- Move mid-imports to top in test files

## Testing

- All unit, integration, and MPI integration tests pass
- New test: `test_vectorized_matches_scalar` — scalar vs vectorized comparison on 10 segments × 5 electrodes (rtol=1e-10)
- New test: `test_point_source_multi_electrode_varying_sigma` — multi-electrode with per-electrode sigma
- New test: `test_coefficients_are_in_mV_per_nA` — explicit unit check
- New reference: `weights_line_source_ref.h5` for line-source integration test
- Public API unchanged (new optional `electrode_chunk_size` parameter with default)


tests/unit/test_weights.py
mgeplf, line 1: why not move the tests of `physics.py` out to a `test_physics.py`?

bluerecording/weights.py
```diff
@@ -771,18 +375,26 @@ def _get_objective_csd_array(
...
     sigma: list[float] | None = None,
     path_to_fields: list[str] | None = None,
     objective_csd_array_indices: list[str] | None = None,
+    verbose: bool = True,
 ) -> pd.DataFrame | None:
     """Compute electrode transfer coefficients from pre-computed positions.
 
-    Dispatches to the appropriate coefficient function based on each
-    electrode's type and returns the concatenated result.
+    Groups electrodes by (type, sigma), computes each group as a batch,
```
mgeplf: how much does `computes each group as a batch` actually save? how many electrodes do we expect? there seems to be a lot of complexity to batch and unbatch them, and I'm not convinced we want that complexity - however, having some numbers could convince me

bluerecording/weights.py
```diff
@@ -809,69 +422,97 @@ def get_weights(
...
+        etype = _get_electrode_type(elec)
+        if etype is ElectrodeType.LINE_SOURCE:
+            line_source_indices.append(idx)
+        elif etype is ElectrodeType.POINT_SOURCE:
+            point_source_indices.append(idx)
         else:
-            electrode_type = electrode.type
+            other_indices.append(idx)
+
+    # --- Batch compute LINE_SOURCE ---
```
mgeplf: these comments, as usual, don't tell me anything that the `if line_source_indices` doesn't already

bluerecording/weights.py
```diff
@@ -809,69 +422,97 @@ def get_weights(
...
+            if electrode_type is ElectrodeType.OBJECTIVE_CSD_SPHERE:
+                coeffs = get_coeffs_objective_csd_sphere(mid_positions, epos, all_epos, radius)
+            elif electrode_type is ElectrodeType.OBJECTIVE_CSD_DISK:
+                coeffs = get_coeffs_objective_csd_disk(mid_positions, epos, all_epos, radius, thickness)
+            elif electrode_type is ElectrodeType.OBJECTIVE_CSD_PLANE:
+                coeffs = get_coeffs_objective_csd_plane(mid_positions, epos, all_epos, thickness)
+
+        elif electrode_type is ElectrodeType.DIPOLE_RECIPROCITY:
+            center = mid_positions.mean(axis=1)
+            coeffs = get_coeffs_dipole_reciprocity(mid_positions, path_to_fields[reciprocity_idx], center)
```
mgeplf: why doesn't `get_coeffs_dipole_reciprocity` calculate the center?

tests/unit/test_weights.py
```diff
@@ -101,13 +107,13 @@ def test_add_coeffs_backwards(tmp_path):
         np.testing.assert_equal(h5[f"electrodes/{POPULATION_NAME}/scaling_factors"][:], expected)
 
 
-def test_get_coeffs_line_source():
+def testget_coeffs_line_source():
```
mgeplf: why no underscore between `test` and `get`?

tests/integration/test_weights.py
```diff
@@ -139,3 +161,49 @@ def test_rat_s1_neurite_types(tmp_path):
...
+    sigma = [0.3, 0.3, 0.3, 0.4, 0.4, 0.3, 0.3]
+
+    # Batched computation (groups by type+sigma, reorders)
+    weights = get_weights(pos_df, cols, electrodes=electrodes, sigma=sigma)
+
+    # Individual computation for verification
+    columns = weights.columns
+    node_ids = np.unique(cols[:, 0])
+
+    from bluerecording.weights import _get_segment_midpts
```
mgeplf: can this be imported at the top?

bluerecording/physics.py
```diff
@@ -0,0 +1,684 @@
...
+from sklearn.decomposition import PCA
+
+
+@dataclass
+class SegmentGeometry:
+    """Precomputed segment geometry for vectorized line-source computation."""
+
+    start_pos: np.ndarray  # (N_line_segments, 3) — segment start positions (µm)
+    end_pos: np.ndarray  # (N_line_segments, 3) — segment end positions (µm)
+    seg_lengths: np.ndarray  # (N_line_segments,) — lengths in meters
```
mgeplf: why is this in `meters` when everything else is in `um`?

bluerecording/physics.py
```diff
@@ -0,0 +1,684 @@
...
+            seg_lengths=seg_lengths,
+            seg_dirs=seg_dirs,
+            is_soma=is_soma,
+            soma_positions=soma_positions,
+        )
+
+
+def precompute_segment_geometry(positions: pd.DataFrame) -> SegmentGeometry:
+    """Precompute segment geometry arrays. Use ``SegmentGeometry.from_positions()`` instead."""
+    return SegmentGeometry.from_positions(positions)
```
mgeplf: is this necessary? I couldn't find the tests for `SegmentGeometry::from_positions` since there is this wrapper

bluerecording/physics.py
```diff
@@ -0,0 +1,684 @@
...
+        The positions DataFrame has a MultiIndex column ``(gid, section_id)``.
+        - If section_id == 0, it's a soma (point source).
+        - If consecutive columns share the same section_id, they form a
+          line-source segment boundary pair.
+
+        Args:
+            positions: DataFrame of segment boundary positions (µm), shape
+                ``(3, N_columns)`` with MultiIndex columns ``(gid, section_id)``.
+        """
+        n_cols = len(positions.columns)
```
mgeplf, line 50: why is this df arranged where the number of columns is unbounded? I find it very strange; it makes the manipulations much harder, imo, as opposed the "traditional" layout.

bluerecording/physics.py
```diff
@@ -0,0 +1,684 @@
...
+
+def _get_line_coeffs(
+    start_pos: np.ndarray,
+    end_pos: np.ndarray,
+    electrode_pos: np.ndarray,
+    sigma: float,
+) -> float:
+    """Compute the line-source coefficient for a single segment.
+
+    All positions are in µm and are converted to m internally.
```
mgeplf: `converted to m internally` why? if the natural scale is `micro`, seems strange to change it

Not sure why this internal detail is described in the docstring

mgeplf (commented):
only made it so far today, I'm very unfamiliar with this code, so not sure how useful I can be

tests/unit/test_weights.py
cattabiani: definitely. I moved the stuff in physics and forgot to move the tests

bluerecording/weights.py
cattabiani: Good point... I agree, the batch/unbatch complexity was too much. Refactoring:

- `get_coeffs_line_source` and `get_coeffs_point_source` each accept either a single electrode `(3,)` or multiple `(N_elec, 3)`. One function, no separate batch variant. The broadcasting happens naturally via NumPy on the electrode dimension.
- geometry parsing (`SegmentGeometry.from_positions`) is called once by the caller and passed in: no repeated work regardless of electrode count.
- memory chunking lives in `get_weights` (the orchestrator), not in physics. I thought about extracting it but it's a 3-line for-loop that slices electrodes into chunks of `electrode_chunk_size` (configurable, default 50) to cap peak memory. James told me we could reach 1k electrodes. At that scale, the `(N_segments, N_elec, 3)` intermediates would blow memory without chunking — but the chunking logic is now trivial and fully separate from the physics math.

**TLDR:** same performance, much less complexity. No batch/unbatch dance anymore.

bluerecording/weights.py
cattabiani: removed some comments

bluerecording/weights.py
cattabiani: yes, done

tests/unit/test_weights.py
cattabiani: old typo that was kept even after the refactoring. Fixed

tests/integration/test_weights.py
cattabiani: fixed the mid-imports. removed

bluerecording/physics.py
cattabiani: we should operate in NEURON units. Fixed

bluerecording/physics.py
cattabiani: no, it is a leftover. removed. Thx for noticing

bluerecording/physics.py
cattabiani: It comes from original get_positions. I agree it is strange. I already segregated the flipping partially. 

This is a substantial change to flip to (N, 3) because it touches:

- _get_cell_positions return shape
- All internal position building (np.tile, np.hstack)
- get_positions assembly
- save_positions (pickled format — backward compat with existing position files)
- SegmentGeometry.from_positions (accessing columns by positions.iloc[:, i])
- get_coeffs_point_source (positions.values.T)
- Objective CSD functions (positions.values - electrode_pos[:, np.newaxis])
-Reciprocity functions
- All test helpers (make_two_section_positions, etc.)
- The _get_segment_midpts function
- The integration tests and MPI tests

cattabiani: open an issue:

https://github.com/openbraininstitute/BlueRecording/issues/92

bluerecording/physics.py
cattabiani: removed. I did a check about units and added a test

bluerecording/weights.py
mgeplf: Thanks.

> James told me we could reach 1k electrodes.

Huh, ok, good to know.

bluerecording/physics.py
mgeplf: np

bluerecording/physics.py
mgeplf: ok, I suspected it was from before. I agree that it doesn't need to be changed here

mgeplf (approved):

PR #84, opened 2026-06-19
Optimize save_weights: rank-ordered layout with collective I/O (3)

### LINKED ISSUE #73: sigma-to-electrode is fragile
## Context

I am referring this this:

https://github.com/openbraininstitute/BlueRecording/blob/379e89f4e91f1271de02736840e61e01efa097f8/bluerecording/weights.py#L784

The ordering guarantee is purely positional by convention — there's nothing in the code that enforces or validates it.

sigma is consumed in electrode-iteration order (the sorted electrode names loop in _compute_electrode_coeffs). The index only advances for LINE_SOURCE and POINT_SOURCE entries. So the caller must supply sigma values in the same order those electrode types appear in the sorted electrode list, skipping ObjectiveCSD/Reciprocity entries. If someone reorders the electrode CSV or inserts a new ObjectiveCSD electrode between two LineSource ones, the mapping silently breaks.

TLDR: the sigma-to-electrode association is implicit and fragile — it relies on the user knowing the electrode iteration order and providing sigma values to match. There's no key-based lookup, no validation, and no error if the list is the wrong length.

We should find a better way

### LINKED ISSUE #79: speed up: line source computation can be vectorized with numpy

### LINKED ISSUE #83: parallel saving can be heavily optimized
even if we can save in parallel from each rank with `mpio` the process produces a scattered save that really slows down IO. We should improve that

cattabiani:
Replace the per-node HDF5 write loop in `_add_data` with a single collective write per MPI rank. Each rank now writes a contiguous block of `scaling_factors` in one operation.

Fix: https://github.com/openbraininstitute/BlueRecording/issues/83
Fix: https://github.com/openbraininstitute/BlueRecording/issues/79
Fix: https://github.com/openbraininstitute/BlueRecording/issues/73

## Changes

- **`_init_weights`**: File layout is now rank-ordered (rank 0's segments first, then rank 1's, etc.) instead of globally GID-sorted. Uses `groupby(sort=False)` to preserve rank-concatenation order.
- **`_add_data`**: Writes `coeffs.values.T` as a single contiguous slice using h5py collective I/O. Falls back to a plain write when h5py lacks parallel support (single-rank / unit tests).
- **`save_weights`**: Uses `MPI_Scan` (`comm.scan`) to compute each rank's row offset. All ranks (including empty ones) participate in collective I/O. Emits a warning if running multi-rank without parallel h5py.
- **`_write_neurite_types`**: Same single-write pattern with collective I/O fallback.
- **MPI integration tests**: Changed to order-independent comparison (per-node data lookup instead of raw array equality).

## Rationale

- The SONATA spec does not require globally sorted `node_ids` in electrode files
- libsonatareport uses the same rank-ordered pattern for reports
- neurodamus reads electrode files with `np.where(node_ids == gid)` — no sorted assumption
- With round-robin GID distribution, the old GID-sorted layout made each rank's data non-contiguous, requiring ~3000 small writes per rank

## Testing

- All unit tests pass
- All 4 MPI integration tests pass with `mpirun -n 2`
- No reference file changes needed (per-node data values are identical)


bluerecording/weights.py
```diff
@@ -166,12 +167,17 @@ def _write_electrode_metadata_to_h5(
...
     with a leading zero.  The result has length ``n_nodes + 1``: entry *i*
     is the index of the first segment for the *i*-th node, and the last
     entry is the total number of segments.
+
+    Uses pandas groupby to preserve appearance order (stable), unlike
+    np.unique which sorts globally.
     """
-    _, counts = np.unique(section_ids_frame["id"].values, return_counts=True)
+    # Group by 'id' in order of first appearance, count segments per node
+    counts = section_ids_frame.groupby("id", sort=False).size().values
```
mgeplf: I though `.values` was not to be used any more (same elsewhere)

bluerecording/weights.py
```diff
@@ -166,12 +167,17 @@ def _write_electrode_metadata_to_h5(
 def _get_offsets(section_ids_frame: pd.DataFrame) -> np.ndarray:
     """Compute per-node offsets into the flat segment array.
 
-    Counts segments per node and returns their prefix sum (partial sum),
+    Counts segments per node in the order they appear (preserving
```
mgeplf: for my understanding, why is `preserving rank-concatenation order` useful?

bluerecording/weights.py
```diff
@@ -166,12 +167,17 @@ def _write_electrode_metadata_to_h5(
...
+    rank-concatenation order) and returns their prefix sum (partial sum),
     with a leading zero.  The result has length ``n_nodes + 1``: entry *i*
     is the index of the first segment for the *i*-th node, and the last
     entry is the total number of segments.
+
+    Uses pandas groupby to preserve appearance order (stable), unlike
```
mgeplf: it's not that it's stable, it's that it's not sorted at all, right?

bluerecording/weights.py
```diff
@@ -221,14 +229,46 @@ def _init_weights(
     """
     comm = MPI.COMM_WORLD
     rank = comm.Get_rank()
+    size = comm.Get_size()
 
-    # Gather all rank-local cols to rank 0
-    all_cols_list = comm.gather(cols, root=0)
+    # Use Gatherv for efficient buffer-based transfer (no pickle overhead)
+    local_count = len(cols)
+    counts = np.array(comm.gather(local_count, root=0))
```
mgeplf, line 229: isn't this using a pickle now vs. comment above?

bluerecording/weights.py
```diff
@@ -627,32 +655,68 @@ def save_weights(
...
     )
+    t1 = MPI.Wtime()
+    log_rank0(f"save_weights: file initialized. ({t1 - t0:.1f}s, includes MPI sync)")
 
-    # 2. Write coefficients in parallel
-    comm = MPI.COMM_WORLD
+    # 2. Compute each rank's contiguous row offset using MPI_Scan
+    local_segments = len(cols)
+
+    if comm.Get_size() > 1 and not h5py.get_config().mpi:
```
mgeplf: shouldn't this be in the `# 3. Open file for parallel write` if?

bluerecording/weights.py
```diff
@@ -627,32 +655,68 @@ def save_weights(
...
-    _add_data(h5, node_ids, weights, population_name)
-
-    if neurite_types is not None:
-        _write_neurite_types(h5, cols, node_ids, neurite_types, population_name)
-
+    comm.Barrier()
     h5.close()
+    t4 = MPI.Wtime()
+
+    # Summary
```
mgeplf: not useful comment imo

bluerecording/weights.py
```diff
@@ -221,14 +229,46 @@ def _init_weights(
...
+        # Build node_ids in rank order: for each rank's cols, extract unique GIDs
+        # (locally sorted within that rank), concatenate in rank order.
+        # GIDs are disjoint across ranks (round-robin distribution).
+        node_ids_parts = []
+        for r in range(size):
+            start_r = int(offsets_per_rank[r])
+            end_r = int(offsets_per_rank[r + 1])
+            if end_r > start_r:
+                node_ids_parts.append(np.unique(all_cols[start_r:end_r, 0]))
+        node_ids = np.concatenate(node_ids_parts) if node_ids_parts else np.array([], dtype=np.int64)
```
mgeplf: perennial why are IDs signed?

bluerecording/weights.py
cattabiani: ah true. I will use `to_numpy()`

bluerecording/weights.py
cattabiani: it is a side-effect. Reworded description

bluerecording/weights.py
cattabiani: no mention of ordering anymore. `sort=False` is already explicit

bluerecording/weights.py
cattabiani: the comment was partially from an older version partially referring to the Gather after. I removed it

bluerecording/weights.py
cattabiani: yes. moved the warning to step 3

bluerecording/weights.py
cattabiani: removed. However, it is more a matter of taste. This function is an orchestrator that does multiple things. I like comments that split logical regions

bluerecording/weights.py
cattabiani: - python/numpy defaults to int64
- libsonata uint64
- neurodamus uint32

I will proceed with uint64 and mention the reasoning there

bluerecording/weights.py
mgeplf: I was thinking of putting it in 679's `comm.Get_size() > 1`

mgeplf (dismissed):

Issue #93, opened 2026-06-30
Clarify and fix output units across all electrode types


cattabiani:
## Problem

The SONATA electrodes_file spec requires scaling_factors in **mV/nA**. Currently, different electrode types may produce inconsistent units:

- **PointSource / LineSource**: formula is `1/(4πσr)` with r in meters and a `× 1e-9` factor. Output is effectively V/nA.
- **Reciprocity / DipoleReciprocity**: formula is `potential / CurrentApplied`. The Sim4Life field files use Dirichlet boundary conditions (1V / -1V) and store the resulting current as `CurrentApplied` (e.g. 415527 in the test file `Infinite_Close_HighRes_SmallSphere.h5`). Dimensional analysis suggests output is V/A, which is `1e-6 mV/nA`. However, empirical comparison with point/line source at the same position shows only ~18× difference, not 1e6×.
- **ObjectiveCSD_***: dimensionless binary masks (0/1).

## Key questions

1. What units does Sim4Life export? The potential field range is [-1, 1] (normalized?), mesh axes are in meters, and `CurrentApplied = 415527` is clearly not 1 A. The paper (S1 Methods) describes Dirichlet BCs of ±1V, with current computed by surface integration of j = σE.
2. What conversion factor is needed on the reciprocity output to produce mV/nA?
3. Should point/line source use NEURON units (µm + S/m → mV/nA directly, no conversion) or SI with explicit conversion?

## Affected tests

- `tests/integration/test_weights.py::test_single_cell_write_weights_near` (uses Reciprocity + DipoleReciprocity + PointSource + LineSource)
- `tests/integration/test_weights.py::test_single_cell_write_weights_distant` (same mix)
- `tests/unit/test_physics.py::test_get_coeffs_reciprocity`
- `tests/unit/test_physics.py::test_get_coeffs_dipole_reciprocity`
- `tests/unit/test_physics.py::test_coefficients_are_in_mV_per_nA`

## Reference files

All `.h5` reference files under `examples/*/reference/` will need regeneration once units are fixed. The current references were generated by the old code and their unit semantics are inherited from that.

## Proposal

1. Adopt NEURON units internally: µm for distances, S/m for sigma, mV/nA for output
2. Point/line source: `1/(4πσr_µm)` gives mV/nA directly — no conversion factors
3. Reciprocity: determine correct factor from Sim4Life conventions (contact Joseph Tharayil)
4. Regenerate all reference files
[...]

PR #94, opened 2026-06-30
Fix point/line source output units: V/nA → mV/nA (1)

cattabiani:
The SONATA electrodes_file spec requires scaling_factors in mV/nA. The formula 1/(4πσr) with r in µm and σ in S/m gives mV/nA directly.

- Remove * 1e-6 position conversions and * 1e-9 current conversion from _get_line_coeffs, _get_coeffs_line_source, _get_coeffs_point_source
- Positions stay in µm throughout (no conversion to meters)
- Update unit test expected values to µm-based formula
- Update reference files (non-mask columns × 1e3)
- xfail single_cell tests (mixed reciprocity types, pending issue #93)

Reciprocity electrode units are unchanged and need separate investigation (see issue #93).


mgeplf (dismissed):
I have to take your word for it

PR #96, opened 2026-07-02
Fix units. Output: V/nA (reciprocity was allright. The spec was wrong)

### LINKED ISSUE #93: Clarify and fix output units across all electrode types

cattabiani:
After the latest development in https://github.com/openbraininstitute/BlueRecording/issues/93 it is clear that the spec and the code must align with V/nA output. We need to update the code

Fix: https://github.com/openbraininstitute/BlueRecording/issues/93


mgeplf:
Should Joe have a look at this?

cattabiani:
definetely! @joseph-tharayil do you want to check this? I cannot ask for a review formally

cattabiani:
actually now this is exactly what we do not want to do. it is simpler to close this and open a new pr

PR #98, opened 2026-07-10
Remove extra trailing column from scaling_factors

### LINKED ISSUE #97: Remove extra trailing column from scaling_factors
BlueRecording writes `scaling_factors` with shape `(Total_comp, N_elec + 1)`, appending a column of ones. The SONATA spec defines the shape as `(Total_comp, N_elec)`. The extra column should be removed to match the spec.

cattabiani:
scaling_factors was written with shape (N_comp, N_elec + 1), appending a column of ones. The SONATA spec defines shape as (N_comp, N_elec). This removes the extra column.

the reference files have dropped the 1 since it is not in the spec

Fixes #97


mgeplf (commented):
does the newly generated files load correctly with neurodamus for simulation?

cattabiani:
@mgeplf I checked and neurodamus was picking just the same so we always had the last electrode with 1 weights (a summation report). I think that was a bug

mgeplf:
> I think that was a bug

I think that should be verified, is there anyone that can look? Maybe @joseph-tharayil?

cattabiani:
@mgeplf he already answered here:

https://github.com/openbraininstitute/BlueRecording/issues/97#issuecomment-4967065262

mgeplf:
Thanks, I missed that.

mgeplf (dismissed):

PR #101, opened 2026-07-13
Fix: write 0-based SONATA node IDs in weights file

### LINKED ISSUE #100: BlueRecording saves gids WITH offsets
The problem is that neurodamus requires them without offset. It is also much more in line with sonata etc. We should store the gids without offset

cattabiani:
Write SONATA node IDs (0-based) instead of neurodamus offset GIDs throughout the pipeline.

- `init_circuit` now returns `(cells, cols, population, population_name, morphologies_dir)` — no more `node_manager`, `ids`, or `gid_offset` in the public API
- `cols` carries 0-based SONATA node IDs from the start (offset subtracted internally)
- `get_positions` accepts cell objects directly instead of `(node_manager, ids)`
- `save_weights` / `_init_weights` no longer need a `gid_offset` parameter
- Updated reference files and all tests

Fix: https://github.com/openbraininstitute/BlueRecording/issues/100


bluerecording/circuit.py
```diff
@@ -23,14 +23,17 @@ def init_circuit(path_to_config: str):
...
+            neurodamus).
         cols: (N, 2) int64 array of (gid, section) pairs describing every
-            compartment on this rank.
+            compartment on this rank. GIDs include the neurodamus offset.
         population: libsonata NodePopulation, needed for morphology file
             resolution.
         population_name: Name of the SONATA node population.
         morphologies_dir: Fully resolved path to the morphologies directory,
             as provided by libsonata.
+        gid_offset: The neurodamus GID offset for this population. Subtract
```
mgeplf: is this parameter needed? Seems like a knob that no one should ever use.

cattabiani: it is used. However, this sparked an internal debate among me, myself and the LLM. I will put this in draft until I find a solution that does not pollute the api

cattabiani: I removed the offset passing because it was leaking an implementation detail. I still need to pass:
- cells
- raw gids

the problem is that I cannot ask for the cell with a raw gid to the node_manager. I need the final gid. Thus, I changed the api slightly and pass around the cell references instead of the ids. It should be very similar space-wise

mgeplf (approved):

bluerecording/circuit.py
mgeplf: ok

PR #102, opened 2026-07-14
Fix output units: mV/nA → V/nA

### LINKED ISSUE #93: Clarify and fix output units across all electrode types

cattabiani:
Point/line source output V/nA (SONATA spec) instead of mV/nA.

- Add 1e-3 factor with unit derivation comments
- Reciprocity already outputs V/nA (unchanged)
- Update references and unit tests

Fix: https://github.com/openbraininstitute/BlueRecording/issues/93


bluerecording/physics.py
```diff
@@ -220,24 +223,30 @@ def _point_source_coeffs(
...
         positions_um: Segment positions (µm), shape ``(N_points, 3)``.
         electrode_positions_um: Electrode positions (µm), shape ``(N_elec, 3)``.
         sigma: Extracellular conductivity (S/m). Scalar or array of shape ``(N_elec,)``.
 
     Returns:
-        Coefficients array (mV/nA), shape ``(N_elec, N_points)``.
+        Coefficients array (V/nA), shape ``(N_elec, N_points)``.
     """
+    # Unit derivation for point source: V = I / (4π σ r)
+    # With σ in S/m and r in µm (= 1e-6 m):
```
mgeplf, line 237: is this LLM output? ` (= 1e-6 m)` doesn't seem particularly useful to me, or things like `has units m/(S·m) = 1/S = Ω`

mgeplf (commented):
I'm still confused as to how this worked in the past if the units were wrong; is there a another PR on the consuming side?

bluerecording/physics.py
cattabiani: no. The LLM actually did not want to add the units. I wanted it for future reference

cattabiani:
@mgeplf 
>I'm still confused as to how this worked in the past if the units were wrong; is there a another PR on the consuming side?

previously it was secretly `V/nA` even if the spec was `mV/nA`. This is to go back to the past and remove as many magic numbers as possible

cattabiani:
@mgeplf I created the problem based on the wrong spec here:

https://github.com/openbraininstitute/BlueRecording/pull/94

mgeplf:
>  based on the wrong spec here:

ahhhh, ok, I finally see.

bluerecording/physics.py
mgeplf: huh, ok, thanks.

mgeplf (dismissed):

PR #107, opened 2026-07-17
Skip connectivity loading during circuit init

### LINKED ISSUE #106: Skip connectivity loading to reduce init time and memory
BlueRecording only needs cell morphologies, compartment structure, and coordinate mappings from neurodamus. It does not use synaptic connections.

Currently, `init_circuit` loads the full circuit including all connectivity, which is expensive for large circuits (time and memory).

Neurodamus supports `restrict_connectivity=2` which skips all connectivity loading (both internal and projections).

Suggested in: https://github.com/openbraininstitute/prod-circuit-simulation/issues/227#issuecomment-5003416437

cattabiani:
Skip all connectivity loading during circuit initialization by passing `restrict_connectivity=2` to neurodamus.

BlueRecording only needs cell morphologies and compartment structure, not synaptic connections. This reduces initialization time and memory usage for large circuits.

Suggested in: https://github.com/openbraininstitute/prod-circuit-simulation/issues/227#issuecomment-5003416437

Fix: #106

### DIFF
--- a/bluerecording/circuit.py
+++ b/bluerecording/circuit.py
@@ -44,6 +44,7 @@ def init_circuit(path_to_config: str):
             sim_config_obj,
             disable_reports=True,
             enable_coord_mapping=True,
+            restrict_connectivity=2,
             simulator="NEURON",
         )
         node_managers = nd.circuits.node_managers


cattabiani:
ok, I see that you are not around today. I will merge by myself

jplanasc (commented):
That's exactly it 😄

